Issue #4, opened 2018-04-08
Add Plex


davestephens:
Add Plex for media transcoding/streaming

Docker container: https://hub.docker.com/r/linuxserver/plex/

Homepage: http://plex.tv

joshmello:
was just going to ask this lol

PR #10, opened 2018-09-23
add plex via @tcharewicz for #4

chrisbutler:
hey @davestephens, i manually cherry-picked the plex work from this fork: https://github.com/tcharewicz/ansible-nas and got it working!

the one issue i ran into is needing to browse to `<nasHostName>:32400/web` to get the plex page to load, which i found out here: https://github.com/ernestasen/ansible-nas/issues/4#issue-311185128


davestephens:
Hey @chrisbutler, this is awesome, thanks for the contribution!

Couple of minor issues however;

- It looks like there are a few undefined variables (`plex_movies_directory`, `plex_tv_directory`) so the playbook won't work as it stands (my fault for not getting the tests working yet :-)). It would actually be better to use `movies_root` and `tv_root` here which are already defined.

- The containers are set to use host networking - meaning we can't remap/change ports of services if we need to. Was there a reason for this? If not, please can you remove those lines and map the specific ports required? 

- Not a massive issue, but plexpy is actually now called tautulli - news to me! If you can be bothered, please can you rename accordingly?

davestephens:
Hey - not forgotten about this - real life has just got in the way. Hope to get this checked out and merged this week! 👍

davestephens:
Finally got round to checking this out - it works great, thanks for your help getting this running! :+1:

chrisbutler:
thank you for publishing it in the first place!

PR #11, opened 2018-10-06
Update all.yml.dist

HariRVA:
Fixes this error:

TASK [Create Nginx Directories] **************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "'nginx_data_directory' is undefined"}


davestephens:
Thank you for this - hope you don't mind I moved your fix to the "shouldn't need to edit" section!

Insert "need to spend time on tests" rant here. :)

Issue #13, opened 2018-11-03
Failure on Package Update


thomascrha:
When I set `keep_packages_updated = true` in my `all.yml` I received this error - 

```
TASK [Upgrade all packages] ******************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unable to find APTITUDE in path. Please make sure to have APTITUDE in path or use 'force_apt_get=True'"}

PLAY RECAP ***********************************************************************************************************************************************************************
localhost                  : ok=15   changed=0    unreachable=0    failed=1
```

SOLUTION:

After then running `apt install aptitude` it worked. So I'm thinking of adding in a condition that if this setting is true check if aptitude is installed, if not install it. I've got no idea if this is doable - I very new to ansible - but I thought this would be a great start - hopefully a PR soon.

davestephens:
What version of Ubuntu is this on? Please can you run `cat /etc/lsb_release` and copy the output?

thomascrha:
```
tcrha@hopenas:~$ cat /etc/lsb_release
cat: /etc/lsb_release: No such file or directory
tcrha@hopenas:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
tcrha@hopenas:~$ which lsb_release
/usr/bin/lsb_release
```

davestephens:
Doh, sorry - the actual file is lsb-release, so it should have been:
```
cat /etc/lsb-release
```
That said I can see you're on 18.04. Still a bit weird though, the module should "just work" - I'll do some digging.

C-J1:
Hey folks. I found this myself when running the playbook on a "stock" Ubuntu 18.04.1 LTS. I added a step to the general tasks to install aptitude first, and that solved it. I'll raise a PR for it (my first ever PR, so please forgive if I screw it up :D)

davestephens:
@C-J1 @thomascrha What ansible versions are you both using? It seems like this used to be a bug but is fixed in newer ansible version. Bug info: https://github.com/ansible/ansible/issues/30516

Please can you run ```ansible --version``` and paste the output?

davestephens:
Fixed in #16

PR #14, opened 2018-11-10
Add install aptitude at start

C-J1:
Added install step for Aptitude right at the start of the general tasks. Aptitude is required for the "Upgrade all packages" step to succeed.


davestephens:
Hey, please see #13 and let me know your thoughts... :+1:

C-J1:
Don't need this one anymore.

PR #18, opened 2018-11-13
Add basic timezone support

C-J1:
Added in the basic code for setting the timezone, and also updated all the containers that I could see that would use it directly. A couple still mount /etc/timezone, and I can't find any info on their pages that they support the TZ environment variable.

I ran the playbook with all of the relevant containers enabled, and they all appear to set the timezone correcctly. (Some you can see in the GUI, others I connected to the console and ran the 'date' command).

I have a follow up to this PR that includes support for setting up NTP. I'll raise that one also.

### DIFF
--- a/group_vars/all.yml.dist
+++ b/group_vars/all.yml.dist
@@ -43,6 +43,11 @@ znc_enabled: false
 # Sets the hostname of your Ansible NAS
 ansible_nas_hostname: bender
 
+# Sets the timezone for your Ansible NAS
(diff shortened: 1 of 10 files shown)


C-J1:
Hey @davestephens , are you looking at accepting this one? I've got a bunch of stuff in the pipeline waiting for this :)

Regards,

CJ

davestephens:
Sorry bro - real life got in the way this week. Awesome addition though, thank you!

PR #19, opened 2018-11-13
Add support for ntp

C-J1:
Added NTP support on top of the basic timezone stuff. This would be my preferred PR to include (I plan to use NTP on my NAS when I actually get around to rebuilding it from FreeNAS. All this has been prep for that. I'm enjoying learning ansible though :D)


davestephens:
Out of curiosity, what are you using this to provide time to?

davestephens:
NTP is a protocol for providing time to things, with the idea being that you have one golden source of time (hence my question what you were using it for). The "usual" service used in Linux is ntpd. It's usually useful for things like messaging, clusters, trading systems, kerberos authentication - anything sensitive to times being synced and accurate, basically.

Where you've set the timezone on containers - that can be done without running an ntpd daemon - all you're doing is telling the container what your local timezone is. 

If you're still keen on a container to run an ntpd that's cool with me - but I think the PR should be separate to one setting the timezones on containers. If you're not fussed about ntpd then we can just remove it from this one and merge accordingly :+1:

davestephens:
Also - if you *are* keen on ntpd, we should run it from a container and not the Ansible role.

C-J1:
I've been using ntpd as a time client. This discussion prompted me to do some research and:

https://help.ubuntu.com/lts/serverguide/NTP.html

You learn something every day. You can reject this PR if you like :)

Regards,

CJ

Issue #3, opened 2018-04-08
Add Gogs


davestephens:
Would like to add Gogs for private Git source control

davestephens:
Gitea appears to be the better of the two projects.

davestephens:
You've hit the exact reason I hadn't progressed with this yet :-)

My ideas were either;

- Document assigning an additional IP to the Ubuntu box, then have a config item of something like gitea_ip_to_use = 192.168.1.31. https://www.garron.me/en/linux/add-secondary-ip-linux.html. I think this would require some extra retro-fitting to tell various services to use the first configured IP.

- The same as your 2.

Thoughts?

C-J1:
That's certainly a possibility, but is it a bit of overkill for what will generally be a home NAS? Then again, changing the host SSH port falls in to that category as well... 

So we have:

- Add multihomed functionality to our host
- Move host SSH to a new port (my preference, just as a good security practice, even though it is just a home NAS)
- Move Gitea to a different SSH port
- Disable SSH access to Gitea, and document that people can only use HTTP(S) for interacting (not a terrible option. Bad, but not terrible)

Say we went with host SSH change. Realistically, how many people are likely to enable gitea? If we document it such that this is an "advanced option", it might be ok. We'd certainly have to ensure that people don't run the playbook on the localhost (restarting SSH). Or that if enabling gitea, you must run playbook A, reconnect to new port, run playbook B (ugly).

I'll have a crack at my preferred option and see how it pans out. I'll raise a PR once I've got what I'd call a working version.

Regards,

CJ

davestephens:
My preference is that we leave SSH on the normal port, and Gitea gets a non-standard port.

Having SSH on a different port would be annoying every time you try to connect to your NAS, whereas Gitea on a different port is likely to not be noticed, given any clone of a repo is a cut and paste of the URL the application gives you...

C-J1:
You have a good point. Also the gitea documentation suggests that moving gitea to a different port is a common practice (https://docs.gitea.io/en-us/install-with-docker/ )

I've got it running, just tweaking some of the options that are available via ENV settings for a cleaner install experience. Have a look if you like before I raise a PR for it.



I'd be interested in knowing how to get Ansible to inject the FQDN of the host...

PR #20, opened 2018-11-13
Add support for gitea. WIP

C-J1:
Not quite ready to be signed and sealed. Raised the PR to share the WIP

Issue #3

### DIFF
--- a/group_vars/all.yml.dist
+++ b/group_vars/all.yml.dist
@@ -31,6 +31,7 @@ stats_enabled: false
 duplicati_enabled: true
 nextcloud_enabled: false
 nginx_enabled: true # if you enable nginx, you'll need to map ports 80 and 443 on your router to your Ansible-NAS
+gitea_enable: false
 
 # IRC bouncer
 znc_enabled: false
--- a/group_vars/all.yml.dist
+++ b/group_vars/all.yml.dist
@@ -31,7 +31,7 @@ stats_enabled: false
 duplicati_enabled: true
 nextcloud_enabled: false
 nginx_enabled: true # if you enable nginx, you'll need to map ports 80 and 443 on your router to your Ansible-NAS
-gitea_enable: false
+gitea_enabled: false
 
 # IRC bouncer
 znc_enabled: false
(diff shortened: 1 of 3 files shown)


group_vars/all.yml.dist
```diff
@@ -31,6 +31,7 @@ stats_enabled: false
 duplicati_enabled: true
 nextcloud_enabled: false
 nginx_enabled: true # if you enable nginx, you'll need to map ports 80 and 443 on your router to your Ansible-NAS
+gitea_enable: false
```
davestephens: Should be ```gitea_enabled```

tasks/gitea.yml
```diff
@@ -0,0 +1,50 @@
...
+
+- name: Create Gitea container
+  docker_container:
+    name: gitea
+    image: gitea/gitea:1.6
+    pull: true
+    links:
+        - gitea-mysql:db
+    volumes:
+      - gitea_data:/data
```
davestephens: Interested on why you've chosen volumes over bind mounts?

davestephens:
This is awesome, thank you! I'll pull and test tomorrow :+1:

tasks/gitea.yml
C-J1: No particular reason. I've done a bunch of reading on volumes vs bind mounts, and everything seems to be "use volumes, they are awesome", except if you want to access the content outside of the container (via, for example, a samba share :D).

With bind mounts, you get the permissions issues, but more flexibility. I'm happy to follow your standard of bind mounts, but I'd like to look at the permissions side of things. Similar to the way that FreeNAS jails handled it (create a app specific user on the host, give it permissions based on what it needs to access (create groups like movieusers or mediausers for example), and then pass in the UID/GID (where appropriate) to the container.

Thoughts?

C-J1:
I swapped to bind mounts for this. It does make sense to be consistent across all the containers.

I hope you don't mind the git user and group accounts. I felt that it makes it a bit easier to allow access to app conf and data directories by letting normal users be added to the group. Something I'd like to replicate across the other containers (eventually). It also stops the apps being run as the root user inside their containers.

davestephens:
The main reason I prefer bind mounts is that you know where your data will actually be on your filesystem, making backups, troubleshooting, general nosiness etc easier!

Sorting the permissions properly has been on my list for ages, just not got to it yet, so this is perfect :+1:

davestephens:
@C-J1 This is awesome dude, thank you! (actually gonna use this one myself too :+1:)

PR #22, opened 2018-11-20
Add support for watchtower

C-J1:
Sorry about the excessive merge commits in here. I did them before I knew better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ just a stock Ubuntu install, some clever Ansible config and a bunch of docker co
   - [Tautulli](http://tautulli.com/) Monitor Your Plex Media Server
   - [Telegraf](https://github.com/influxdata/telegraf) - Metrics collection agent
   - [Transmission](https://transmissionbt.com/) BitTorrent client (with OpenVPN if you have a supported VPN provider)
+  - [Watchtower](https://github.com/v2tec/watchtower) Monitor your Docker containers and update them if a new version is available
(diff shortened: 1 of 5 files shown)


davestephens:
Have a read of this - you can squash your commits into one sane one ;-)

http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

davestephens:
So this is an interesting one. I'm currently working on a feature to pin containers to a specific image version so that upgrades are a bit more controlled. Bear in mind that a new image version can often equal a completely new software version with breaking changes - not cool for stability or someone that wants it to "just work" because they're fed up of FreeNAS (ie, me). After running ansible-nas for a year or so multiple containers were updated and subsequently busted by these types of upgrades - transmission-with-openvpn, grafana and Portainer come to mind. 

That said, I get that there are people that want the latest and greatest as soon as they are available so I *am* happy to include Watchtower - I just don't think it should be enabled by default.

C-J1:
Sure, I can see the the desire for stability over all. Definitely an important feature on the "just work" front.

That said, there are options to manage which containers get auto updated  or not:

`--label-enable Watch containers where the com.centurylinklabs.watchtower.enable label is set to true.`

And no problems about not having it enabled by default. I'll switch that around and try to squash. I might have to delete my remote repo to do it, so not sure what will happen to this PR :D

C-J1:
Closing this due to new, improved, clean history PR #24

PR #24, opened 2018-11-20
Added watchtower support

C-J1:
Re-created the branch to clean up the excessive number of commits :)

### DIFF
--- a/group_vars/all.yml.dist
+++ b/group_vars/all.yml.dist
@@ -323,26 +323,17 @@ guacamole_data_directory: "{{ docker_home }}/guacamole"
 ###
 ### Watchtower
 ###
-# Sets the 6 field cron schedule to use for checks and updates
-watchtower_cron_schedule: "0 0 5 * * *"
+# Sets the 6 field cron schedule to use for checks and updates. This will check at 5am daily.
+watchtower_cron_schedule: 0 0 5 * * *
 
-## Notifications
-# If enabling notifications, you MUST correctly setup the relevant
-# section below
-watchtower_notifications_enabled: false
-# Set the type of notification. email, or slack
-# uncomment below if you wish to use notifications
-watchtower_notifications_type: email
-# Set the notifications level. panic, fatal, error, warn, info (default) or debug
-watchtower_notifications_level: info
+# Sets the Watchtower Docker start command. Different options can be supplied based on whether you want to receive
+# notifications or not, some examples are provided below. See https://github.com/v2tec/watchtower for more info.
 
-# Email
-watchtower_email_from: "{{ ansible_nas_email }}"
-watchtower_email_to: "{{ ansible_nas_email }}"
-watchtower_email_server: smtp.gmail.com
-watchtower_email_server_user: "{{ ansible_nas_email }}"
-watchtower_email_server_password: abcd1234
+# No notifications
+watchtower_command: "--schedule '{{ watchtower_cron_schedule }}' --debug"
 
-# Slack
-watchtower_slack_hook_url: https://hooks.slack.com/services/???/???/???
-watchtower_slack_identifier: "watchtower-{{ ansible_nas_hostname }}"
+# Email notifications
+# watchtower_command: "--schedule '{{ watchtower_cron_schedule }}' --notifications 'email' --notification-email-from 'ansible@nas.com' --notification-email-to 'notifications@ansible-nas.com' --notification-email-server 'my.email.server.com' --notification-email-server-port '25' --notification-email-server-user 'email_username' --notification-email-server-password 'top-secret'"
+
+# Slack notifications
+# watchtower_command: "--schedule '{{ watchtower_cron_schedule }}' --notifications 'slack' --notification-slack-hook-url 'https://hooks.slack.com/services/xxx/yyyyyyyyyyyyyyy' --notification-slack-identifier 'ansible-nas'"
--- a/group_vars/all.yml.dist
+++ b/group_vars/all.yml.dist
@@ -34,6 +34,7 @@ glances_enabled: true
 stats_enabled: false
 guacamole_enabled: false
 netdata_enabled: false
+watchtower_enabled: false
 
 # Backup & Restore
 duplicati_enabled: true
@@ -65,6 +66,37 @@ ansible_nas_email: me@example.com
 
 # Applications will have subdomain SSL certificates created, eg ansible-nas.<your-domain>, nextcloud.<your-domain>
 ansible_nas_domain: example.com
+
+###
+### Watchtower
+###
+# Sets the 6 field cron schedule to use for checks and updates
+# Keep in mind that this will stop and start containers if there is an update
+# So probably best not to do it every 10 minutes unless you want your plex stream
+# interrupted
+# Default below is every day at 5am (local time if you have set
+# your local timezone above)
+watchtower_cron_schedule: "0 0 5 * * *"
+
+## Notifications
+# If enabling notifications, you MUST correctly setup the relevant
+# section below
+watchtower_notifications_enabled: false
+# Set the type of notification. email or slack
+watchtower_notifications_type: email
+# Set the notifications level. panic, fatal, error, warn, info (default) or debug
+watchtower_notifications_level: info
+
+# Email
+watchtower_email_from: "{{ ansible_nas_email }}"
+watchtower_email_to: "{{ ansible_nas_email }}"
+watchtower_email_server: smtp.gmail.com
+watchtower_email_server_user: "{{ ansible_nas_email }}"
+watchtower_email_server_password: abcd1234
+
+# Slack
+watchtower_slack_hook_url: https://hooks.slack.com/services/???/???/???
+watchtower_slack_identifier: "watchtower-{{ ansible_nas_hostname }}"
 
 ###
 ### Docker
--- a/group_vars/all.yml.dist
+++ b/group_vars/all.yml.dist
@@ -333,7 +333,7 @@ watchtower_cron_schedule: 0 0 5 * * *
 watchtower_command: "--schedule '{{ watchtower_cron_schedule }}' --debug"
 
 # Email notifications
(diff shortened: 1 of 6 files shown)


C-J1:
Recreated this based on what I did leading up to PR #22 

Cleaner history, etc

davestephens:
Thanks for this. I'm still struggling to get my head around how this could interact with pinning docker images to specific versions.

Think of this scenario:

Grafana image pinned in the playbook to 5.3.4
Watchtower enabled upgrades it to 5.4
Ansible-nas playbook runs again and spins up 5.3.4
Watchtower sees its not up to date, and upgrades....

etc etc...Any ideas of how this could be worked around if watchtower was enabled?

C-J1:
I'd suggest using the watchtower label feature:

`--label-enable Watch containers where the com.centurylinklabs.watchtower.enable label is set to true.`

I haven't tested, but I'd assume that if the label is not specified, the container is not watched. I'm not sure how you'd do this in ansible, but something like:

if <container>_lockversion: 1.2.3 then set label to false? or don't set label at all?
else if <container>_autoupdate_enabled: true set label to true
else do nothing

Something like that anyway. I'll have dig around and see if I can figure out how you'd do that without having 3 different "create container" steps (which watchtower has currently anyway).

Regards,

CJ

C-J1:
Ran a test overnight last night. I specified an older version of radarr during install, and left watchtower to run. It did not auto update radarr. Debug logs were informative:

```
Checking containers for updated images
Retrieving running containers
Pulling v2tec/watchtower:latest for /watchtower
No credentials for v2tec in /config.json
No authentication credentials found for v2tec/watchtower:latest
No new images found for /watchtower
Pulling netdata/netdata:latest for /netdata
No credentials for netdata in /config.json
No authentication credentials found for netdata/netdata:latest
No new images found for /netdata
Pulling linuxserver/nzbget:latest for /nzbget
No credentials for linuxserver in /config.json
No authentication credentials found for linuxserver/nzbget:latest
No new images found for /nzbget
Pulling linuxserver/radarr:139 for /radarr
No credentials for linuxserver in /config.json
No authentication credentials found for linuxserver/radarr:139
No new images found for /radarr
Pulling linuxserver/sonarr:latest for /sonarr
No credentials for linuxserver in /config.json
No authentication credentials found for linuxserver/sonarr:latest
No new images found for /sonarr
Pulling portainer/portainer:latest for /portainer
No credentials for portainer in /config.json
No authentication credentials found for portainer/portainer:latest
No new images found for /portainer
Scheduled next run: 2018-11-27 06:00:00 +1100 AEDT
```

Tonight I'll try and see how it handles incremental updates to a specific tag (ie/ release-15. Will it update if it sees release-15.01)

davestephens:
Ah, so I guess it just looks for updates to the tag of the image that's already running, which means it'd work nicely. I did look at the docs too and it wasn't overly clear how it works or what it's supposed to do, other than "update images". 

With regards to your changes, please can you tidy them up a bit? Specifically the all.yaml.dist - there's more watchtower stuff in there than pretty much everything else combined - I don't think it's necessary to support every possible Watchtower config option (Slack support is way overkill, for example) :) 

Will also tap some comments into the changes :+1:

.gitignore
```diff
@@ -1,2 +1,3 @@
 group_vars/all.yml
 group_vars/vpn_credentials.yml
+roles/
```
davestephens: Please can you drop this change from this PR?

group_vars/all.yml.dist
```diff
@@ -59,6 +60,37 @@ ansible_nas_user: david
...
+###
+### Watchtower
+###
+# Sets the 6 field cron schedule to use for checks and updates
+# Keep in mind that this will stop and start containers if there is an update
+# So probably best not to do it every 10 minutes unless you want your plex stream
+# interrupted
+# Default below is every day at 5am (local time if you have set
+# your local timezone above)
+watchtower_cron_schedule: "0 0 5 * * *"
```
davestephens: All of this cron stuff can be moved down to the "you won't need to edit this" section. A 5am nightly update is sufficient for 99% of use cases IMO

group_vars/all.yml.dist
```diff
@@ -59,6 +60,37 @@ ansible_nas_user: david
...
+# Email
+watchtower_email_from: "{{ ansible_nas_email }}"
+watchtower_email_to: "{{ ansible_nas_email }}"
+watchtower_email_server: smtp.gmail.com
+watchtower_email_server_user: "{{ ansible_nas_email }}"
+watchtower_email_server_password: abcd1234
+
+# Slack
+watchtower_slack_hook_url: https://hooks.slack.com/services/???/???/???
+watchtower_slack_identifier: "watchtower-{{ ansible_nas_hostname }}"
```
davestephens: I can't think of any good reason to include the slack support? It's incredibly unlikely I'll have a chat room with me and my NAS. :smile:

group_vars/all.yml.dist
```diff
@@ -59,6 +60,37 @@ ansible_nas_user: david
...
+# If enabling notifications, you MUST correctly setup the relevant
+# section below
+watchtower_notifications_enabled: false
+# Set the type of notification. email or slack
+watchtower_notifications_type: email
+# Set the notifications level. panic, fatal, error, warn, info (default) or debug
+watchtower_notifications_level: info
+
+# Email
+watchtower_email_from: "{{ ansible_nas_email }}"
```
davestephens: this doesn't need to be parameterised, or if you think it does, it should be in the "you don't need to change this" section

group_vars/all.yml.dist
```diff
@@ -59,6 +60,37 @@ ansible_nas_user: david
...
+watchtower_email_to: "{{ ansible_nas_email }}"
```
davestephens: same for this

group_vars/all.yml.dist
```diff
@@ -59,6 +60,37 @@ ansible_nas_user: david
...
+watchtower_email_server: smtp.gmail.com
```
davestephens: I think it makes sense to have an outbound email server/username/password setting bit in the general section, it shouldn't be specific to watchtower, and perhaps suggest mailgun/sendgrid/whatever

group_vars/all.yml.dist
```diff
@@ -59,6 +60,37 @@ ansible_nas_user: david
...
+# Default below is every day at 5am (local time if you have set
+# your local timezone above)
+watchtower_cron_schedule: "0 0 5 * * *"
+
+## Notifications
+# If enabling notifications, you MUST correctly setup the relevant
+# section below
+watchtower_notifications_enabled: false
+# Set the type of notification. email or slack
+watchtower_notifications_type: email
```
davestephens: not necessary if slack is ditched

group_vars/all.yml.dist
```diff
@@ -59,6 +60,37 @@ ansible_nas_user: david
...
+# Set the notifications level. panic, fatal, error, warn, info (default) or debug
+watchtower_notifications_level: info
```
davestephens: move to "you probably don't need to change this" section

tasks/watchtower.yml
```diff
@@ -0,0 +1,57 @@
+---
```
davestephens: I don't think we need three different variations of one container. Just one with email support is fine, and if the options aren't supplied the emails will simply go nowhere. I think. Thoughts?

group_vars/all.yml.dist
C-J1: Ironically, this is my go-to notification system. I have a dedicated slack workspace for all sorts of notifications. My current setup has sonarr, radarr, plex (tautulli) and emby all send notifications to slack channels. You can configure netdata to send slack alarms, etc. I'm going to integrate slack support for SMART alerts, etc. I'm more likely to see a slack notice in a timely manner than I am an email :)

group_vars/all.yml.dist
C-J1: I parameterised them so that people had the option of changing who the emails come from and go to. ie/ my current freenas setup has emails (where used) coming from a nasmon@... email address.

My preference is to have good, solid defaults "above the line" which a decent amount of flexibility "below the line".

group_vars/all.yml.dist
C-J1: That makes sense. I'll change it tonight

tasks/watchtower.yml
C-J1: This is one I struggled with. I wanted the flexibility to support all the notification methods supported (including none) in just one task, but couldn't figure out how to make it work.

I'd be afraid that if email details aren't supplied, it would just generate error spam in the container, which would be problematic if a logging system were implemented (ELK stack or the like).

davestephens:
Y'know what, following your comments I've let this one percolate a bit today and I actually quite like the idea of a Slack room where all of the different containers post their notifications to. It's certainly a lot cleaner than tonnes of email. 

So back on the topic of Watchtower....perhaps we go with no notifications or slack. Thoughts?

C-J1:
I think the options need to remain. We can't expect everyone (anyone) to setup email, slack, or msteams (I should add that in, but don't want to setup teams just to test it) just to have notifications. I think the default should still be notifications = false, but maintain the below the line options to enable.

So the PR as is I think?

C-J1:
If you really wanted to get creative, you could create a container for mattermost and use that in place of slack :) I'm pretty sure that mattermost supports slack compatible webhooks...

davestephens:
> If you really wanted to get creative, you could create a container for mattermost and use that in place of slack :) I'm pretty sure that mattermost supports slack compatible webhooks...

Yeah...no :-)

I do like the idea of being opinionated though - i.e. notification are through Slack, or not at all. I don't use email at work as it's a pain in the ass, most comms is over Hipchat. It's also something a bit different - pushing the user to not do things the "old" way.

C-J1:
Where are we at with this one @davestephens ? :)

davestephens:
Hmm...the commit history of this ended up a bit messier than I'd hoped! :-/

Hope you don't mind - I've added a doc page, and also made it so the config is more easily configurable to do exactly what a user might want without repeating the container declaration multiple times in the `tasks/watchtower.yml` file. 

Merge conflict aside, what's your opinion on this approach?

C-J1:
> Hope you don't mind - I've added a doc page, and also made it so the config is more easily configurable to do exactly what a user might want without repeating the container declaration multiple times in the `tasks/watchtower.yml` file.
> 
> Merge conflict aside, what's your opinion on this approach?

It's good. I prefer the environment variable approach for readability, but this solves the multi container definition well.

davestephens:
Good to get this one merged, thanks for your patience with getting this sorted :+1:

Issue #23, opened 2018-11-20
files/folders in roles/ directory causing problems in GitKraken


C-J1:
The roles/samba and roles/docker directories are causing me some issues. GitKraken (on Windows) thinks they are supposed to be submodules, but I can't see a .git directory in there at all.

Regards,

CJ

davestephens:
Doh, I see what's happened here. I used to load in the roles via git submodules but changed it a while back  to use ansible-galaxy instead. Looks like my last commit accidentally added part of the submodule config back in :man_facepalming:

davestephens:
Should be sorted.

PR #25, opened 2018-11-23
Fix issue #23

C-J1:


.gitignore
```diff
@@ -1,2 +1,3 @@
 group_vars/all.yml
 group_vars/vpn_credentials.yml
+roles/
```
davestephens: This line needs to be removed or it'll be impossible to add roles in the future

davestephens:
Thanks for this - is a mistake from my last commit to add Guacamole! Will need the gitignore change reverting then I can merge it in.

davestephens:
Should be fixed in commit c3712fa

PR #26, opened 2018-11-26
Add support for multiple download clients

C-J1:
Added in SABnzbd as NZB download client.

To achieve this, I felt it a good idea to change how the downloads directory structure was setup. This will allow other download clients in the future (NZBGet for example) as alternatives.

Regards,

CJ


davestephens:
This is cool, is something that's been on my list for a while! 

Am on the train so difficult to read the diff right now, but it looks like the watch directory for Transmission and SABnzbd are the same, which would mean it's not possible to run both at the same time, and have them both watch for torrents. I also think (looking for guidance here!) it's a legit use case to have both SABnzbd and Transmission active at the same time?

group_vars/all.yml.dist
```diff
@@ -44,6 +46,7 @@ znc_enabled: false
 ###
 # Sets the hostname of your Ansible NAS
 ansible_nas_hostname: bender
+ansible_nas_domain: homenetwork
```
davestephens: Please can you remove this change?

C-J1:
A bit of work to do in here I think...

In order for all the download handlers (torrent, nzbget, sabnzbd, etc) to work smoothly with the media management tools (radarr, sonarr, etc) then the paths inside the containers need to match across the containers. With some of the tools (sonarr and radarr) you can do some remote path mapping stuff, but if we get it right at the "physical" path level, then things should just "work".

For example:

application | path on host | path in container
------------ | ------------- | -------------
transmission | /mnt/volX/downloads/transmission/complete | /downloads/transmission/complete
nzbget | /mnt/volX/downloads/nzbget/completed | /downloads/nzbget/completed
sonarr | /mnt/volX/downloads | /downloads

In the example, the idea is that sonarr (and other tools of a similar nature) will talk directly to the download clients via API and, once a download is finished, will read the path to the completed download ([ref](https://github.com/Sonarr/Sonarr/wiki/Completed-Download-Handling)). If that path doesn't match across the containers, then sonarr or radarr would not find the file. We could overcome that with remote path mappings (for apps that support it), but I feel that would add an additional level of complexity (and not work in all cases), considering that technically all of the hosts are "nas@domain" with port variations.

And this has also got me thinking longer term regarding media locations. ie/ someone mentioned airsonic... and that'd be great, but we'd then have to add a /audiobooks mount in to all the relevant download client configs. Better to just have something like a /media path (which we kind of have with "samba_shares_root"), so that again all the paths would match across all of the containers.

Regards,

CJ

C-J1:
Added some more changes here. Still a work in progress, but I wanted some more eyes looking at it :)

C-J1:
Closing this PR in favour of doing it a bit differently. The end goal is absolutely the same.

PR #28, opened 2018-11-26
Set docker up on dedicated ZFS dataset

C-J1:
I'm not sure if this is something you had considered doing. I'm going to implement this when I rebuild my FreeNAS box, so I figured it couldn't hurt to make a PR for it.

### DIFF
--- a/group_vars/all.yml.dist
+++ b/group_vars/all.yml.dist
@@ -38,7 +38,6 @@ gitea_enabled: false
 znc_enabled: false
 
 
-
 ###
 ### General
 ###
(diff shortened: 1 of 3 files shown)


davestephens:
Thanks for this @C-J1 ! 

What's the thinking behind further embedding ZFS into Ansible-NAS? Is there any benefit to doing so or any major performance gains? I personally find it (ZFS) a pain in the ass, and its support on Linux is certainly not as bullet-proof as BSD variants, for example. Although I've never bothered to document it (or anything for that matter, but that does need to change!) - ZFS support in ansible-nas in my eyes has always been an enabler to moving away from it to better things, not using it more and further complicating configuration. 

That said, if there genuinely is a good reason for this change (rather than just because we can) then I'm open to being persuaded...

C-J1:
To me a NAS has always been about reliable data storage whilst minimizing the potential for data loss. I'm not aware of a filesystem other than ZFS that delivers that as well as ZFS does. Granted there are cons to using it (high RAM requirement, inability to remove a drive, to name a couple). ZFS is one of the main reasons my current NAS is FreeNAS (I'm close to migrating, not there yet). I'm primarily moving away because I want docker without running a VM to do it (I tried for awhile, but things were soooo slow).

And I think it is a good idea to have a "product" settle on a recommended way of doing things. You say having support for it is an enabler to move to better things... what better things? I'd be really interested to know, and if it's good, help you build it.

Regards,

CJ

C-J1:
I can just create the relevant datasets manually before I run the playbook. That's probably the better implementation, as it retains that filesystem flexibility.

davestephens:
OK. I think that actually makes sense - I don't want to get into automating filesystem creation or destruction as getting it wrong makes you very unpopular :)

PR #29, opened 2018-11-27
Add support for netdata

C-J1:
Add support for netdata

Includes fix that allows docker container names to be visible


inventory
```diff
@@ -1,3 +1,3 @@
 [all]
-192.168.1.30
+nas-test.thecave ansible_python_interpreter=/usr/bin/python3
```
davestephens: Please can you remove this from the PR?

C-J1: Done

davestephens:
This is awesome, thank you. Sorry for the delay in getting it checked out!

PR #34, opened 2018-12-23
Remove inventory and make inventory.dist file.

jgoerz:
- This file changes for each location.


davestephens:
Nice idea, thanks for the contribution!

PR #40, opened 2018-12-27
Update syntax to remove deprecation warning

jgoerz:
- Warning:
TASK [Install some packages] ***************************************************
[DEPRECATION WARNING]: Invoking "apt" only once while using a loop via          
squash_actions is deprecated. Instead of using a loop to supply multiple items 
and specifying `name: "{{ item }}"`, please use `name: ['smartmontools',        
'htop', 'zfsutils-linux', 'bonnie++', 'unzip', 'lm-sensors']` and remove the 
loop. This feature will be removed in version 2.11. Deprecation warnings can be 
 disabled by setting deprecation_warnings=False in ansible.cfg.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,9 @@ inventory
 roles
 *.retry
 *.log
+
+# Vim
+[._]sw[a-p]
+
+# Session
+Session.vim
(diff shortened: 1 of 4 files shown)


davestephens:
I think it'd be cleaner to move the list of packages out to `all.yml` (and `all.yml.dist` and `tests/test.yml`) and then reference the variable rather than in-line a list of packages. There's actually an example in the Ansible 2.7 porting guide that goes something like this:

```
- name: Install packages
  apt:
    name: "{{ ansible_nas_extra_packages }}"
    state: present
```

PR #42, opened 2018-12-30
Sync test.yml with latest vars

jgoerz:
- Added watchtower_enabled and its settings.
- Update extra packages.


jgoerz:
This is to prevent ```fatal: [default]: FAILED! => {"msg": "The conditional check 'watchtower_enabled' failed....``` errors.  I do not know how to fix the build so it catches these types of omissions.

davestephens:
Thanks for this, I need to think of a clever way to fail the build so this situation doesn't occur again!

Issue #35, opened 2018-12-26
Update Traefik to using letsencrypt staging env by default


C-J1:
When building and rebuilding containers that will reach out to letsencrypt for certs, we can very quickly hit the rate limit for production certs.

It is better to, by default, use the letsencrypt staging environment. This ensures that any testing and build processes being followed don't hit the prod rate limits. When you are ready to do a production run of your ansible playbook, you can just change the setting to use prod environment, delete your traefik container and let it run. This would assume that router port forwarding is setup and working (define test method in doco?)

Documentation and a "How To" for this would be required. Even if you run without firewall port forwarding, you can still consume rate limits if you aren't careful.

davestephens:
I guess smashing the LetsEncrypt rate limit is techincally a bug as it's undesired...have tagged accordingly!

C-J1:
I rebased this on to current master. @davestephens  Have a look when you can

jesse-troy:
I'm very interested in this. Since it's been about a month, I figured I would bump to see if you guys have had any testing/success with this. I am unable to test this currently, but have been following this issue closely for a while.

C-J1:
It works just fine. I finally migrated from FreeNAS over the weekend, and used this to ensure I was getting certs properly.

davestephens:
On Kubernetes this is a pain in the bum - you have to manually delete the
staging certs before switching to the production ones.

Did you have to do this here it was it automatic?

C-J1:
I manually deleted the acme.json file.

I can add some logic where we have a flag file for which mode it is in. Then if we see a switch between prod and staging, we can delete the acme file (backing up first of course).

davestephens:
Ok cool.

It'd be nice if we could automate that, you know what the first issue will
be after flicking to staging certs by default :-)

PurpleNinja225:
prolly can be closed as switching between staging and prod servers is as easy as editing ```traefik_acme_server:``` (Unless the manually deleting acme.json is still a requirement during the switch) I haven't tested this exactly yet, but I haven't encountered this issue yet in 2.X

davestephens:
It can. I tested it a week or so ago and you still need to delete it, but it's documented here: https://github.com/davestephens/ansible-nas/blob/75103190ce4fb5c9187b8992ef41964e1a1729f3/roles/traefik/defaults/main.yml#L25

PR #44, opened 2019-01-03
Fix issue #35 - Traefik requests production cert by default

C-J1:
First cut at fixing issue #35. I'm not able to test this properly at home at the moment. I should be able to do some proper testing on the weekend, but I think this is ok. If someone can test sooner, that'd be awesome.

### DIFF
--- a/docs/applications/traefik.md
+++ b/docs/applications/traefik.md
@@ -9,9 +9,59 @@ for each application in the Advanced Settings section of your `all.yml`.
 
 See [External Access](configuration/external_access) for more info.
 
-## Usage
+## Let's Encrypt (SSL), Staging and Production environments, and rate limits
+
+A bit of background so that the usage section becomes easier to understand.
+
+All secure websites on the internet have a certificate that is issued by a trusted certficate
+issuer (eg/ GoDaddy, Verisign, etc). The website certificate is "signed" by what is called the "root" certificate of the issuer. Your computer and web browser has a list of these "trusted"
+root certificates, so that when you visit a secure website, your browser can say "this website's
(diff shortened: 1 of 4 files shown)


davestephens:
This is awesome, definitely needed, thank you! :+1: 

I reckon we can make it a little cleaner such that the container won't need a manual restart or deletion when you switch from staging to production though:

* Change `traefik.toml` to `traefik.toml.j2` and update the template step
* Add a conditional block around the staging parameter in the `traefik.toml.j2`, something like:
   ```
   {% if traefik_production == false %}
     caServer = "https://acme-staging-v02.api.letsencrypt.org/directory"
   {% endif %}
   ```
* Add a conditional restart to the traefik docker container when the template step is changed (see how docker is restarted after the config changes for an example)

Also, is it worth documenting the difference between staging and production LetsEncrypt certs?

C-J1:
Have a fresh look at the doco. I also force pushed and squashed some commits to make the branch a bit cleaner.

C-J1:
@davestephens ping! :)

davestephens:
I vaguely recall a conversation we had about removing the received cert between switching from staging -> production - did anything get done with that - or am I imagining it?

C-J1:
Yeah we did have that convo. I've been focusing on Real Life (tm) lately :D

I'll get to it as soon as I can.

Regards,

CJ

davestephens:
That's fine, I just couldn't find where we had the conversation so was
checking I'd not imagined it! I might sort it this week if I get a chance,
assuming that's ok with you?

C-J1:
Of course

davestephens:
Now supported in Traefik Ansible role 👍

Issue #47, opened 2019-01-05
Playbook assumes universe repo is enabled


C-J1:
**Short problem description**

The playbook assumes that the universe repository is enabled. This is fine on a local host run where the installation instructions say to enable the repo (needed to install ansible itself). This shouldn't be necessary for running on a remote host

**Environment**
- Ansible-NAS revision (`git rev-parse --short HEAD`): b7d94d2
- Ansible version (paste the entire output of `ansible --version` on the machine you run the playbook from): 
```
ansible 2.5.1
  config file = /home/craig/source/ansible-nas/ansible.cfg
  configured module search path = [u'/home/craig/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.15rc1 (default, Nov 12 2018, 14:31:15) [GCC 7.3.0]
```
- Ansible-NAS operating system (`uname -a` on the Ansible-NAS box): 
`Linux utest 4.15.0-43-generic #46-Ubuntu SMP Thu Dec 6 14:45:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`
- Ansible-NAS Python version (`python --version` on the Ansible-NAS box):
```
craig@utest:~$ python --version

Command 'python' not found, but can be installed with:

sudo apt install python3       
sudo apt install python        
sudo apt install python-minimal

You also have python3 installed, you can run 'python3' instead.
```

**Expected behavior**

Playbook should run through smoothly with no errors during the "install some packages" task.

**Actual behavior**

Playbook fails to install the "lm-sensors" package.

**Steps to reproduce**

Run the playbook on a fresh server that does not have the universe repo enabled 

**Playbook Output**
```

Using module file /usr/lib/python2.7/dist-packages/ansible/modules/packaging/os/apt.py
<nas-test.thecave> ESTABLISH SSH CONNECTION FOR USER: None
<nas-test.thecave> SSH: EXEC ssh -C -o ControlMaster=auto -o ControlPersist=60s -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o ConnectTimeout=10 -o ControlPath=/home/craig/.ansible/cp/6473105f95 nas-test.thecave '/bin/sh -c '"'"'echo ~ && sleep 0'"'"''
[...]

PR #48, opened 2019-01-05
Resolves issue #47

C-J1:
Add general task to enable universe repo


C-J1:
I had tried to do this via the apt_repository module, but that wants a specific deb command. So it would be difficult to handle mirror's, different releases, etc. This seemed the best way to achieve it.

PR #49, opened 2019-01-05
Some basic defaults changes

C-J1:
A suggestion for some basic defaults changes. Reasoning inline

* Traefik - disabled by default
  * Not everyone is going to need or want external access to their containers.
* airsonic - disabled by default
  * I'm guessing this was a mistake when airsonic was added
* duplicati - disabled by default
  * Not everyone will want / be able to use the backup service. It almost certainly warrants some investigation before enabling and using it
* keep packages updated - enabled by default
  * Good practice to keep your system updated
* Moved portainer to first installed container
  * I encountered a situation where traefik container broke (whilst I was working on another PR), and I wanted to delete it. I used the docker CLI, but it did occur to me that it would be great if container management was there from there start.

### DIFF
--- a/group_vars/all.yml.dist
+++ b/group_vars/all.yml.dist
@@ -7,7 +7,7 @@
 # Traefik will allow access to certain applications externally. To enable this you'll need a domain name that points to your 
 # home static IP address, or use a dynamic DNS provider like no-ip. You'll also need to map ports 80 and 443 from your router
 # to your ansible-nas server.
-traefik_enabled: true
+traefik_enabled: false
 
 # BitTorrent
 # If you plan to use Transmission with OpenVPN, you'll need to copy group_vars/vpn_credentials.yml.dist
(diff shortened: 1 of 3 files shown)


davestephens:
Agree with all of this except the package updates. It means the playbook can take AGES to run, and some people may not want grub automatically updating for them, for example. Agree it's a good thing to have, but it should be opt-in rather than forced. 

Maybe we should investigate properly configuring unattended upgrades as part of the playbook instead?

C-J1:
> Agree with all of this except the package updates. It means the playbook can take AGES to run, and some people may not want grub automatically updating for them, for example. Agree it's a good thing to have, but it should be opt-in rather than forced.
> 
> Maybe we should investigate properly configuring unattended upgrades as part of the playbook instead?

Yep, fair enough. I've turned it back off again and pushed up the change.

Sounds like a good idea, but I've no idea where to start on that one :)

C-J1:
One I forgot to include in this. Add a true/false control for Heimdall. Have it enabled by default, but at least have it as a choice.

If you're good with that, I'll add it in.

davestephens:
Yep definitely, makes sense!

PR #51, opened 2019-01-13
Add missing variable for test.yml

jgoerz:
Fix error when running tests/test-vagrant.sh:
Unable to find any of pip2, pip to use.  pip needs to be installed.


davestephens:
Doh, thanks for this! FYI @C-J1 :-)

PR #55, opened 2019-02-23
Add new doc section with overview for NAS newbies

scotws:
Adds a new section to the documentation for the inexperienced with links to get them started. Hope this will help people decided if this is the right project for them. You'll probably want to tweak the text a bit in some places, like aims.

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -1,0 +1,85 @@
+Ansible-NAS currently assumes you know your way around a server. This page is an
+overview for absolute NAS beginners so they can decide if it is right for them.
+
+## The big picture
+
+To start off _really_ simple: A NAS ([Network Attached
+Storage](https://en.wikipedia.org/wiki/Network-attached_storage)) is a server
+mostly for home or other small networks that offers file storage. It's usually a
+small box that sits in the corner and runs 24/7. These days, a NAS doesn't just
+only handle files, but also offers other services, for instance video streaming
+with [Plex](https://www.plex.tv/) or [Emby](https://emby.media/index.html). You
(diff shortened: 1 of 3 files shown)


docs/overview.md
```diff
@@ -0,0 +1,85 @@
...
+
+With Ansible-NAS, at this point at least, you're pretty much on your own. Though
+there is a [Gitter](https://gitter.im/Ansible-NAS/Chat) chat room (see
+[support](support.md)), you're expected to have some familiarity with the
+technologies involved and be able to set up the basic stuff yourself. 
+
+As a to-do list, before you can even install Ansible-NAS, you'll have to:
+
+1. Choose, buy, configure, and test your own **hardware**. Note that ZFS eats
+   RAM like a pig, prefers to have the hard drives all to itself, and should be
```
davestephens: I think it may be worth clarifying ZFS memory requirements as it depends massively on the use case, or perhaps linking something for people to reference. ie, I run my Ansible-NAS on a poxy old AMD with 16gb memory, and it's fine, even with tonnes of Docker containers running.

Same for ECC memory - is this relevant for the home user? It depends on your use case :-) I happen to agree with you, but is someone that's download torrents and saving tv shows to disk going to care?

davestephens:
This is awesome, and fills a big glaring gap in the docs, thank you. Have added a comment around the hardware bit, aside from that I think we're good to merge. :+1:

docs/overview.md
scotws: Good point. I'll add some more stuff on that.

davestephens:
Perfect, thank you!

PR #56, opened 2019-02-25
Deluge

tcharewicz:
Add deluge application to ansible-nas repository.


davestephens:
Thank you for submitting this PR, getting Deluge merged will be awesome. This has however, reminded me why I didn't do it myself in the first place :smile: 

Few things that will need fixing;

- You've supplied a default username/password, but not mentioned in the docs what it is or how to change it (I think mentioning the default for the container, and how to change it is fine)
- The docs need to mention that having Transmission and Deluge pointing at a watch directory will cause problems
- The paths in the supplied config don't look right (ie `/root/Downloads`)
- The added variables need adding to `tests/test.yml`
- You've specified a local network range that's different to the one specified in the example config and transmission. It's probably worth pulling this (and the transmission one) out to a central variable in the config shared by both containers.

Also I'm not clear on;
- Why are config files supplied? Is there something you're changing that's required that isn't possible to do via environment variables? (if so, is there a better image floating around that could be configured with environment variables?)

tcharewicz:
Please tell me, why you didn't do this?

Deluge after installation is not configured properly, it misses configuration about watch folder and where to download files that why I add configuration files to the repository. But I can delete these files, and add information to deluge docs the user after the first login needs to configure watch and downloads directory manually in configuration from webUI.

davestephens:
Have a read of this: https://github.com/davestephens/ansible-nas/blob/master/docs/contributing.md

It's generally considered good practise to read contribution guidelines before contributing to a project on GitHub. If you don't and still submit a PR, you should expect some sort of feedback along the lines of what the guidelines say, which is what I did. 

I'm not at a computer right now but I want to test what you say about directories properly, if possible I don't want to supply config files. Reason being, if you run the playbook, change the config, run the playbook again then you're going to break people's config for them, which is not great.

PR #58, opened 2019-02-26
First post-install instructions

scotws:
Adds the first version of a text about what to do when installation was successful. A lot depends on what the user has chosen to include, but I'm figuring Heimdall should be the way to go one way or another.

### DIFF
--- a/docs/post_installation.md
+++ b/docs/post_installation.md
@@ -5,11 +5,14 @@ dashboard of your new NAS, because most of the applications included come with a
 web interface. Heimdall lets you create "apps" for them which appear as little
 icons on the screen. 
 
-To add applications to Heimdall, you'll need the IP address of your NAS. If you
-don't know the address for some reason and have not installed
-[Avahi](https://www.avahi.org/) for zero-configuration networking (mDNS), you
-can `ssh` into your NAS and type `ip a` to find it. The entry "link/ether",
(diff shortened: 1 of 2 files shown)


docs/post_installation.md
```diff
@@ -0,0 +1,26 @@
...
+The first thing to do is to configure [Heimdall](https://heimdall.site/) as the
+dashboard of your new NAS, because most of the applications included come with a
+web interface. Heimdall lets you create "apps" for them which appear as little
+icons on the screen. 
+
+To add applications to Heimdall, you'll need the IP address of your NAS. If you
+don't know the address for some reason and have not installed
+[Avahi](https://www.avahi.org/) for zero-configuration networking (mDNS), you
+can `ssh` into your NAS and type `ip a` to find it. The entry "link/ether",
+usually the second one after the loopback device, will show the address.
```
davestephens: Isn't this chicken and egg? How can you ssh to it to find the ip address if you don't know the ip address?

davestephens: Also, if you've managed to run Ansible-NAS and don't know the IP address or hostname of your box, then you probably have bigger problems than finding how to get to Heimdall :)

scotws: Uh, yeah, that doesn't sound right. Let me look at that again this weekend.

davestephens:
This is great, thank you! :+1:

PR #59, opened 2019-03-02
Change networking for emby container to host

burrowsa:
This allows devices such as smart TVs to find and stream from emby


davestephens:
I don't have a way to test this so will trust you on this one ;-)

PR #61, opened 2019-03-02
Add mymedia for alexa

burrowsa:


davestephens:
This is awesome, thank you - fixed up a few minor issues so I can merge it, hope that's ok!

burrowsa:
gr8, sorry about the mistakes, took the PR from my own setup.

I found that I needed host networking to make it work, maybe there is another way to squash the warning?

davestephens:
No probs.

The warning was due to both specific ports and host networking being
specified on the container definition.

I followed the docs by removing host_networking as I don't have an Alexa to
test with. Happy to flip it to the other instead if it doesn't work, but
seems odd that the official docs would be wrong?

burrowsa:
In the config webpage for mymediaforalexa you can override the IP but it is simplest to use host mode and let it automatically work out the IP. I think it is probably best to remove the port.

davestephens:
Switched it back!

PR #67, opened 2019-03-22
Adding Jackett

jesse-troy:
I have a lot of features I want to add that I have running on my own system, but I am not at all familiar with using git (or github) so I am submitting this very simple feature first.

Jackett assists media sourcing applications (sonarr, radarr, etc) with searching torrent trackers. 

Please let me know if I should be doing this any differently. I will also message you directly on gitter.

### DIFF
--- a/nas.yml
+++ b/nas.yml
@@ -132,5 +132,5 @@
     tags: get_iplayer
 
   - import_tasks: tasks/jackett.yml
-    when: jackett_enabled
+    when: (jackett_enabled | default(False))
     tags: jackett
--- a/nas.yml
+++ b/nas.yml
@@ -130,3 +130,7 @@
   - import_tasks: tasks/get_iplayer.yml
     when: (get_iplayer_enabled | default(False))
     tags: get_iplayer
+
+  - import_tasks: tasks/jackett.yml
+    when: jackett_enabled
+    tags: jackett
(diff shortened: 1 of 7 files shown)


nas.yml
```diff
@@ -130,3 +130,7 @@
   - import_tasks: tasks/get_iplayer.yml
     when: (get_iplayer_enabled | default(False))
     tags: get_iplayer
+
+  - import_tasks: tasks/jackett.yml
+    when: jackett_enabled
```
davestephens: Please can you make the `when:` clause include a default as with the other blocks

davestephens:
This is awesome, thank you!

Only thing missing is adding it to the README, and a minor tweak required on the nas.yml. 

Aside from that, good to go :+1:

nas.yml
jesse-troy: @davestephens I have made this change, but I don't know how to update this pull request. Could you point me in the right direction?

jesse-troy: I have also updated the README with the following line, but for some reason it is not showing up here. 
 `* [Jackett] (https://github.com/Jackett/Jackett) - API Support for your favorite torrent trackers`

Git/Github is very confusing to me for some reason.

davestephens: Looks good to me - you can see all of the changes in the diff view: https://github.com/davestephens/ansible-nas/pull/67/files

davestephens:
Merged! 😎

PR #69, opened 2019-03-25
Add documentation page for Emby

scotws:
Since we have two media servers, Plex and Emby (actually, three if you count Jellyfin, which should be trivial to add), we might want to add a line or two at some point explaining why there are in fact that many, that is, what the differences are (Plex wants you to register, Emby isn't completely open source any more, etc). For the moment, this is just a basic note on the app itself.


docs/applications/emby.md
```diff
@@ -0,0 +1,107 @@
...
+connect to from various platforms - other computers, smartphones, smart TVs,
+etc.
+
+Note that [Plex](https://www.plex.tv/), also included in Ansible-NAS, has a very
+similar functionality.
+
+## Usage
+
+Set `emby_enabled: true` in your `group_vars/all.yml` file. There are further
+parameters you can edit such as `movies_root` and `tv_root` lower down in this
```
davestephens: Not sure the info about extra params is relevant to this doc page?

docs/applications/emby.md
```diff
@@ -0,0 +1,107 @@
...
+file. 
+
+## Specific Configuration
+
+The emby web interface can be found at port 8096 (http) or 8920 (https, if
+configured) of your NAS. After installation, point your browser there to start
+the configuration wizard. For instance, on a NAS with the IPv4 address
```
davestephens: Not sure we need to teach people how to browse to a url ;-)

docs/applications/emby.md
```diff
@@ -0,0 +1,107 @@
...
+192.168.1.2, browse to `http://192.168.1.2:8096`. Heimdall has a dedicated icon
+for emby. 
+
+By default, Ansible-NAS gives emby read/write access to the folders where your
```
davestephens: One thing I'm not willing to encourage people to do is edit the task files - primarily because it makes pulling latest changes difficult WRT to merge conflicts. I don't think there's any value in setting a volume to RO, but would accept a PR to move the config out to the all.yml if you think there's value in it.

docs/applications/emby.md
```diff
@@ -0,0 +1,107 @@
...
+      - "{{ emby_movies_directory }}:/movies:rw"
+      - "{{ emby_tv_directory}}:/tv:rw"
+```
+
+so that they end in `ro` instead of `rw`. Note that emby will not be able to
+delete files then, which might be exactly what you want. However, you will not
+have the option to store cover art in the related folders. Always leave the
+configuration directory read/write. 
+
+## File system considerations
```
davestephens: This is great info, but not specific to Emby. Perhaps could be moved out to a ZFS page?

docs/applications/emby.md
```diff
@@ -0,0 +1,107 @@
...
+can't be compressed and aborts the procedure. 
+
+> At time of writing, native encryption of ZFS datasets is not supported in the
+> release versions of ZFS on Linux (ZoL), but is scheduled for [version
+> 0.8.0](https://github.com/zfsonlinux/zfs/releases/tag/zfs-0.8.0-rc3) sometime
+> in 2019. It is unclear when Ubuntu will upgrade their version of ZFS.
+
+To see all properties of a ZFS dataset, use `zfs get all <DATASET>`.
+
+## Naming movies and TV shows
```
davestephens: There must be a better way of doing this than telling people to rename all video files?

davestephens:
As always, thank you for the awesome docs! Have added few comments that I'd appreciate your thoughts on. :+1:

docs/applications/emby.md
scotws: (Agree on the text changes above) The idea is to make absolutely sure that neither Plex nor Emby nor anybody else gets the idea that they can delete my media files (and that's just me being paranoid, go over to /r/datahorder on Reddit and check out their stuff). Of course they're RO on a file level, but still. -- Give me few days for the PR, will move it to all.yml.

docs/applications/emby.md
scotws: You're right, we should probably have a "What you need to know about ZFS for Ansible NAS" page somewhere, including a setup for snapshots. I'll open an issue for that and move stuff over there.

(Is it just me or has interest in ZFS on Linux exploded since Ubuntu got behind it? I'm also seeing  a lot more stuff about Proxmox suddenly, but that might just be the company I keep)

docs/applications/emby.md
scotws: (Laughs hysterically) You'd think so in the age of AI, wouldn't you? Sadly, from my experience with Plex, later moved to Emby, no -- you can either name the files the way the system wants it or you can manually match them later through the interface.

The problem, to be fair, is that too much stuff has the same name. Take "The Avengers", which means Scarlett Johansson in leather for some people and Diana Rigg in leather for others. Then of course there is https://www.imdb.com/title/tt0034639/?ref_=fn_tt_tt_7 from 1942, though I'm not sure if anybody is wearing leather. I stopped counting on IMDB at 20 films with the name "Dracula". You get the picture.

scotws:
I'll put this on the back burner for now until I've figured out that PR for moving the configuration to all.yml. It probably also makes more sense to write some ZFS documentation first so we can reference it from here.

scotws:
I'll rewrite this once #79 is included.

scotws:
@davestephens Okay, now that we've moved the ZFS stuff out, let's have another go at this.

davestephens:
:boom:

Issue #74, opened 2019-04-02
Vagrant test script fails due to version mismatch


scotws:
**Short problem description**

After changes to config files, attempt to run test version with `tests/test-vagrant.sh` as specified in README.md produces the following error message in a slightly scary red font:

> This Vagrant environment has specified that it requires the Vagrant
> version to satisfy the following version requirements:
> 
>   >= 2.2.2
> 
> You are running Vagrant 2.0.2, which does not satisfy
> these requirements. (...)

Then the Vagrant VM cannot be accessed. This is on stock Ubuntu 18.04.02 LTS with Vagrant as installed with `sudo apt install vagrant`.

**Environment**
- Ansible-NAS revision (`git rev-parse --short HEAD`): 417f997
- Ansible version (paste the entire output of `ansible --version` on the machine you run the playbook from):

```
ansible 2.5.1
  config file = /home/scot/Programming/ansible-nas/ansible.cfg
  configured module search path = [u'/home/scot/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.15rc1 (default, Nov 12 2018, 14:31:15) [GCC 7.3.0]
```
- Ansible-NAS operating system (`cat /etc/lsb-release` on the Ansible-NAS box) - _If this is anything other than Ubuntu 18.04 help will be limited_:

```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.2 LTS"
```
- Ansible-NAS kernel (`cat /etc/lsb-release` on the Ansible-NAS box): (as above)
- Ansible-NAS Python version (`python --version` on the Ansible-NAS box): (as above)
- Ansible-NAS Docker version (`docker --version` on the Ansible-NAS box): Docker version 18.09.2, build 6247962
- Latest Docker logs (`journalctl -u docker.service` on the Ansible-NAS box): n/a and too long
- Are you running the playbook from a remote box or the Ansible-NAS box itself?  Box itself
- Vagrant version, if testing (`vagrant --version`): Vagrant 2.0.2

**Expected behavior**

Create virtual machine for testing.

**Actual behavior**

Error message as above, VM won't run.

**Steps to reproduce**

[...]

PR #76, opened 2019-04-10
Add bitwarden service

jesse-troy:
Bitwarden is a password management tool, similar to LastPass.

This version of bitwarden is in a single docker container as opposed to the official version which creates something like 6 containers. It also includes a backup tool that will automatically create a backup at 5 a.m. every day. This tool can be configured as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ just a stock Ubuntu install, some clever Ansible config and a bunch of Docker co
 ### Docker Containers Used
 
 * [Airsonic](https://airsonic.github.io/) - catalog and stream music
+* [Bitwarden_rs](https://github.com/dani-garcia/bitwarden_rs) - Self-Hosting port of password manager
 * [Cloudflare DDNS](https://hub.docker.com/r/joshuaavalon/cloudflare-ddns/) - automatically update Cloudflare with your dynamic IP address
 * [CouchPotato](https://couchpota.to/) - for downloading and managing movies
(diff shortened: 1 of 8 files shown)


davestephens:
Ok...so just pulled this and tried to run it up on the Vagrant test Ansible-NAS and it doesn't appear to be working. Any ideas? How do I test it so that I know it's good to merge? :)

scotws:
@davestephens Are you having the same issue as #74 ? That's some weird version mismatch, but I don't understand Vagrant enough to do much more than point out it isn't working for me either.

davestephens:
@scotws Nah Vagrant is working fine, I'm not *that* stupid ;-)

davestephens:
OK so I got my brain into gear and did some more testing on this - it works fine (obvs) but not when spun up within Vagrant. Seems to be the extra hop across into Vagrant screws it up - but whatever - if you `vagrant ssh` and `curl` the web port, everything is listening as expected. 

I've pushed some tweaks which *should* fix the WebSocket thing too, so gonna test that later this evening...

I also fixed the mega merge smash that was my own doing, and have merged your work. Thanks for the contribution! :+1:

jesse-troy:
Okay great! In my testing, it seemed ot work best without websockets (I thoght I had it commented out, my bad). Websockets really only allows "instant" synchronization with web extensions. I was going to work on it some this week, but it looks like you found the fix, to use the traefik protocol!

Issue #73, opened 2019-03-29
Add a "ZFS Intro for Ansible-NAS users" page


scotws:
Ansible-NAS currently assumes that the user will know how to setup and configure ZFS, including snapshot management. It would be nice to at least point out that a program such as Sanoid (https://github.com/jimsalterjrs/sanoid/) can be used to automate the snapshot part of the NAS. 

(Possibly this could be included in a "default ZFS scenario" for people not so familiar with these things as a reference where to start: Install root on "normal" ext4 file system, create pool assuming two mirrored hard drives, set up datasets for TV, Samba shares, movies, etc, set up snapshots of said datasets.)

davestephens:
Never heard of it but this looks awesome, will check it out!

scotws:
As in discussion for #69 , include the Sanoid stuff in an intro page documenting the basics of ZFS, including suggestions on what parameters to use. Might be worth waiting for ZoL 0.8 to be released for this because it includes cool stuff like TRIM support and native at-rest encryption. Should include example for snapshots etc. for most common use case.

scotws:
Unless somebody has a better idea, I would suggest splitting this into two documents: One a general introduction to ZFS as related to the use in NASes, including the ability to move pools from FreeNAS to Ansible-NAS (and, of course, the reverse way); the other specific configuration suggestions and examples for the apps included.

davestephens:
I think we need to consider the target audience too. A lot of people likely
won't care about most of the ZFS options, and just want to know that their
data is mirrored on a pair of disks and will be recoverable. I for one fall
into that camp.

Me aside, there are always power users that want to go "full enterprise" on
their home stack, and that's cool - but I've always been pretty open about
Ansible-NAS *not* managing disks/partitions because a- it's hideously
complicated and b- it's near-on impossible to automate for every use case.

I also don't believe there should be differing ZFS partition configs for
different applications - for the same reasons above, and also that not
everyone uses ZFS.

Any ZFS stuff should be set system-wide and used by everything - let's not
forget this is a home NAS, not an enterprise grade storage system where
disk latency of utmost importance :-)

Guiding new users through creating a sensible ZFS setup, and giving power
users ideas of options to tweak is absolutely the right thing to do, but I
think the documentation needs to call those differing use cases out
clearly.

Tl;Dr; I think we're agreeing, I just want to be clear that none of this
stuff is ever going to be automated, and I don't want to be recommending
loads of manual tweaking every time a new container is spun up.

[...]

PR #79, opened 2019-04-13
Add general and specific ZFS documentation

scotws:
As discussed in #73 , this includes a general intro to ZFS and a specific example for a most very simple Ansible-NAS setup. The one missing part is at the end of the specific introduction, the instructions for sanoid snapshot management, because I haven't done that yet myself and won't get around to it for a while. I suggest we leave the Docker parts as a link to the Docker website so we don't have to change stuff here when they change.


scotws:
Okay, @davestephens I think the build is failing because you're busy with stuff and I'm trying to play catch-up :smile: . Since I'm just off night shift I'm probably going to be awake longer than you are, so I'll wait a few hours and rebase this (and the emby stuff) on your new master then.

davestephens:
> Okay, @davestephens I think the build is failing because you're busy with stuff and I'm trying to play catch-up  . Since I'm just off night shift I'm probably going to be awake longer than you are, so I'll wait a few hours and rebase this (and the emby stuff) on your new master then.

Yeah sorry, that was me breaking the build. If you rebase it should now pass :)

docs/overview.md
```diff
@@ -48,40 +48,27 @@ technologies involved and be able to set up the basic stuff yourself.
...
-   recommended](https://github.com/zfsonlinux/zfs/wiki/FAQ#do-i-have-to-use-ecc-memory-for-zfs)
-   to use ECC RAM. ZFS also prefers to have the hard drives all to itself. If
-   you're paranoid (a good mindset when dealing with servers), you'll probably
-   want an uninterruptible power supply (UPS) of some sort as well and SMART
-   monitoring for your hard drives. See the [FreeNAS hardware
-   requirements](https://freenas.org/hardware-requirements/) as a guideline.
+   RAM, and it is [recommended](zfs_overview.md) you use ECC RAM. ZFS also
+   prefers to have the hard drives all to itself. If you're paranoid (a good
+   mindset when dealing with servers), you'll probably want an uninterruptible
+   power supply (UPS) of some sort as well and SMART monitoring for your hard
```
davestephens: Interested in what you mean by "as well and SMART monitoring for your hard drives"?

docs/overview.md
```diff
@@ -48,40 +48,27 @@ technologies involved and be able to set up the basic stuff yourself.
...
+   drives. See the [FreeNAS hardware
+   requirements](https://freenas.org/hardware-requirements/) as a guideline, but
+   remember you'll also be running Docker.
 
 1. Install **Ubuntu Server**, preferably a Long Term Support (LTS) edition such
    as 18.04, and keep it updated. You'll probably want to perform other basic
```
davestephens: Ubuntu needs to be 18.04 - this is not an optional thing.

docs/overview.md
davestephens: I also think a home NAS project recommending ECC is way overkill. I run Ansible-NAS on an old HP Microserver. I don't do enough io that ECC memory is going to save me from any sort of data-loss.

docs/overview.md
```diff
@@ -48,40 +48,27 @@ technologies involved and be able to set up the basic stuff yourself.
...
-   video](https://www.youtube.com/watch?v=MsY-BafQgj4) that introduces the
-   philosophy and big picture of ZFS.
-   
+1. Install **ZFS** and set up storage. You can use a different file system and
+   volume manager, but Ansible-NAS historically tends towards ZFS. You'll have
+   to create datasets for various parts of the system, some form of automatic
+   snapshot handling, and possibly automatic backups to another server or an
+   external hard drive. If you are completely new to ZFS, expect a brutal
+   learning curve. A [brief introduction](zfs_overview.md) is included here, as
+   well as a [basic example](zfs_configuration.md) of a very simple ZFS setup.
```
davestephens: So this is an interesting one and not something that occurred to me on first review. Few things that come to mind: 

- If someone has installed Ubuntu, then they have a root filesystem already, likely ext4, so their requirement is for an *additional* filesystem for storage, and "You'll have to create datasets for various parts of the system" is not strictly true. 
- ZFS packages are installed by Ansible-NAS for the purpose of migrating from FreeNAS and for using as the storage filesystem. There's no reason (that I can think of) where it would need to be installed manually.
- I'm not a fan of the negativity in "If you are completely new to ZFS, expect a brutal learning curve". I learnt the basics required in about 30 minutes, it ain't that hard, and I don't want people scared off before they even start :)

docs/overview.md
```diff
@@ -91,6 +78,6 @@ Ansible-NAS.
 The easiest way to take Ansible-NAS for a spin is in a virtual machine, for
 instance in [VirtualBox](https://www.virtualbox.org/). You'll want to create
 three virtual hard drives for testing: One of the actual NAS, and the two others
-to create a mirrored ZFS pool. Note because of the RAM requirements of ZFS,
-you might run into problems with a virtual machine, but this will let you
-experiment with installing, configuring, and running a complete system. 
+to create a mirrored ZFS pool. A virtual machine will probably not be happy or
```
davestephens: A virtual machine will be perfectly happy, it's just not recommended for "proper" use....because it's a VM. :)

docs/zfs_configuration.md
```diff
@@ -0,0 +1,232 @@
+This text deals with specific ZFS configuration questions for Ansible-NAS. If
+you are new to ZFS and are looking for the big picture, please read the [ZFS
+overview](zfs_overview.md) introduction first.
+
+## Just so there is no misunderstanding
+
+Unlike other NAS variants, Ansible-NAS does not install, configure or manage the
+disks or file systems for you. It doesn't care which file system you use - ZFS,
+Btrfs, XFS or EXT4, take your pick. Nor does it provides a mechanism for
+external backups, snapshots or disk monitoring. As Tony Stark said to Loki in
```
davestephens: It does provide a mechanism for external backups - Duplicati.

davestephens:
The ZFS docs are *awesome*. Thank you so much. They definitely deserve their own area in the docs though - please can you move them to a subfolder of `ZFS`? 

Also, regarding ZFS pools and datasets - rather than saying "create a pool" then "you *could* create X dataset for Y", do you reckon we should just recommend what people set up as a best practise for getting started (or better still, script it/Ansible it?!) Ie a dataset for movies, tv, downloads, dump etc etc? You have a much better understanding on ZFS than I, and I wonder for the new user that just wants to be told what to do - you/we provide that?

docs/overview.md
davestephens: I've just read more of the docs and realised that you say what I say regarding ECC, so we should be consistent.

docs/overview.md
scotws: I'm rewriting the overview text to leave the choice of "specialized bulk storage" more open, point out Btrfs (which at some point in the future is bound to get its RAID5 problems solved, and then should see an uppick in popularity) is an option, so we're more agnostic towards file systems. I cut out the ECC and VM parts here completely, because that can be taken care of in the ZFS texts themselves.

scotws:
I'd hesitate with giving more detail about ZFS, because it can turn very technical very quickly - if you have two 8 TB, drives, a mirror is fine; if you have six 2 TB drives, you start getting into discussions about three mirrors vs two RaidZ1 VDEVs with three drives each or all six drives in a single RaidZ2 VDEV. 

**Suggestion**: We keep one simple, imperative-style "do this" configuration example for ZFS, and consider adding a _third_ text at some point for "advanced configuration". That could involve discussing the trade-offs of mirrors vs RaidZ1 vs RaidZ2 etc, and the use of L2ARC cache drives for virtual machines and/or containers (which you don't need for straight-out movie file serving, so I've ignored it for the moment). At that point, we hopefully have more feedback from users about these text, including what we're missing (or screwed up, cough).

scotws:
BTW, the Ubuntu people seem to be working on a ZFS installer (see https://www.phoronix.com/scan.php?page=news_item&px=ZFS-GRUB-More-Bits-April-Ubuntu), so 20.04 LTS should make at least some of this redundant.

docs/overview.md
utdrmac: I think a comment should be made that while ECC ram is best, make sure your processor and chipset support it. I bought ECC ram not knowing this and my i3 doesn't support ECC. Wasted $50 on memory I cannot use.

scotws: Good point, will update.

davestephens:
Finally got around to reading through this. It's *awesome* - thank you!

PR #80, opened 2019-04-14
Add permissions parameters for Plex and Emby

scotws:
User can now set parameters to decide if their Plex and/or Emby apps should consider the TV and movies folders as read-only or read-write (see discussion at #69). Default is read-only as best practice. 

(Apols, vagrant test won't work on my system (see #74), so this might need further testing by somebody who actually knows what he's doing.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,13 +112,10 @@ Assuming that your Ubuntu system disk is separate from your storage (it should b
 
 ## Requirements
 
-<<<<<<< HEAD
-* Ansible NAS targets the latest Ubuntu LTS release, which is currently Ubuntu Server 18.04.2 LTS. 
-* You can run Ansible-NAS on whatever you like, read the docs for more info. I use an HP Microserver.
-=======
-Ansible NAS should work on any recent Ubuntu box. Development is done on Ubuntu
-Server 18.04.1 LTS running on an HP Microserver.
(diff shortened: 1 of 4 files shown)


davestephens:
Just so I'm clear - this is just to prevent Emby and Plex inadvertently modifying your media? Are there examples where Plex/Emby have done this or is it just a "best practise" thing? I've run both of them historically and they've never fiddled with my media unless I've specifically told them to. 

What I want to avoid is people being tripped up by having their media directories set to `ro` by default, then wondering why Emby and/or Plex aren't working as they expect them to - then later realising Ansible-NAS is to blame. If this is to be included I feel the default should be `rw` with documentation notes about why you *might* want to set the volumes to `ro`, and the pros/cons of doing so. 
 
If set to ro - how do Emby and Plex behave? Are logs filled with errors about being unable to write to media directories or do they just carry on as normal?

Also - your diff seems to include some files you're not changing, can you take a look at that too please?

scotws:
Your right, default for normal, adjusted, non-paranoid people should be `rw`, I'll change it, and then later mention it in the docs. -- The changes to README.md seem to be line breaks only, I tend to have them, because vim, and you seem not to. It doesn't make a difference with Markdown, the docs are a happy mix by now; if you have a preference, I can change it.

davestephens:
:+1:

Issue #81, opened 2019-04-15
Switch to using updated sickchill container


l3rady:
The Sickrage container used is no longer supported and have moved over to a new container called sickchill: https://github.com/linuxserver/docker-sickrage/pull/45

Would be nice to see ansible nas move over to this container instead: https://github.com/linuxserver/docker-sickchill

davestephens:
More than happy to switch and/or accept a PR if you want to contribute 👍

l3rady:
Sure thing: https://github.com/davestephens/ansible-nas/pull/82

PR #82, opened 2019-04-15
Sickrage to sickchill

l3rady:
https://github.com/davestephens/ansible-nas/issues/81


davestephens:
This is awesome, thank you :+1: 

Only thing to consider is if people are currently using Sickrage, then pull this update and run again - the playbook will likely fail due to Sickrage and Sickchill trying to map the same port.

In the `sickchill.yml` please can you add a step to ensure that the Sickrage container is not running before starting up Sickchill? From memory you can run `docker_container` with `state: absent` which should sort it.

l3rady:
@davestephens while in the task file I noticed one of the volumes on the container for completed downloads is set like so: https://github.com/l3rady/ansible-nas/blob/f8554f3be558dace1cebb6b043cf919c8892e612/tasks/sickchill.yml#L23

Don't think this is correct to force the user to a folder ending in `/completed` I for one like to put tv downloads in a folder called `completed-tv`.

Would I be right in saying that the `/completed` part of this line should be moved into the variable `sickchill_downloads_directory`?

davestephens:
> Don't think this is correct to force the user to a folder ending in /completed I for one like to put tv downloads in a folder called completed-tv.

> Would I be right in saying that the /completed part of this line should be moved into the variable sickchill_downloads_directory?

Not quite. `sickchill_downloads_directory` is the same as the Transmission download directory. Transmission creates two directories in its download directory - `completed` and `incomplete` So, Sickchill is configured to pick stuff up the completed directory. 

Bear in mind Sickchill doesn't actually download anything, it just pokes other applications to do the downloading, ie Transmission.

davestephens:
Tested and works perfectly, thank you!

l3rady:
@davestephens thanks for merging. Regarding the completed folder bit:

I understand that sick chill doesn't do the downloads but you can configure, when talking to transmission, to offer a different download directory. In my case, I tell sick chill to use the folder completed-tv and couchpotato to use completed-films that way each program are not scanning each other's downloads folders. Also what if you are not using transmission but something else to download and doesn't use the completed folder scheme. Hence why making the whole path configurable I think is better.

PR #85, opened 2019-04-28
quick sonarr & radarr docs

GREEB:
I just wrote down what i found out for the setup of these two applications

### DIFF
--- a/docs/applications/sonarr-radarr.md
+++ b/docs/applications/sonarr-radarr.md
@@ -1,0 +1,24 @@
+# Sonarr & Radarr
+Homepages: [sonarr](https://sonarr.tv/) / [radarr](https://radarr.video/)
+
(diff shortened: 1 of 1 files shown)


davestephens:
Awesome, can never have too many docs - huzzah! 

Please can you separate this into separate files for Radarr and Sonarr? Appreciate a lot of it is repeated, but cos a person uses Radarr, it doesn't necessarily mean they'll use Sonarr too - so makes sense (to me) to separate.

Also, interesting what you say about the path mapping - had no idea that was necessary as I pointed it at an already existing directory. Please can you provide specific "go here, do this" instructions under a "Specific Config" heading as everyone will no doubt have this issue.

We could just add a step to the radarr/sonarr tasks to create a directory within the directory, which should solve the problem from what you're saying. What do you think?

GREEB:
Hey, sure i can split this up for each app. I'll do some more test to be sure i'm not lying ;)

And yeah putting an empty folder in the /tv & /movies could fix that, let me know how you'd add it so i can also test that.

davestephens:
Superceded by #213

PR #90, opened 2019-05-04
Order all.yml, closes #89

### LINKED ISSUE #89: Reorganize all.yml
The number of entries in the lower parts of all.yml is getting large enough that finding things is becoming hard. Starting about line 243 - the part the user shouldn't touch anyway - we should sort them by name to make everybody's life easier. (I should be able to do a PR later today)

scotws:
Ordering starts after line 243.

### DIFF
--- a/group_vars/all.yml.dist
+++ b/group_vars/all.yml.dist
@@ -241,28 +241,97 @@ ansible_nas_extra_packages:
 
 ansible_python_interpreter: /usr/bin/python3
 
-###
-### Samba
-###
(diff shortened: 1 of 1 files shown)


caillou:
I would not recommend reordering of the lines. This will generate unnecessary merge conflicts when updating to the latest version of `ansible-NAS` and I don’t think it will significantly solve the problem of managing the file over time.

@scotws would you agree about the merge conflicts introduced by this?

PR #93, opened 2019-05-17
Extend ZFS documentation: Snapshots, sanoid

scotws:
Adds the missing sections on snapshots and using sanoid to create and manage them. This should complete the ZFS documentation for now.


davestephens:
This is awesome, thank you. I wonder if it's worth seeing if we can do some sort of recommended Sanoid configuration that could configure an automated setup if people choose to use it...

scotws:
What we could certainly consider is a standard media ZFS dataset ("share") for Plex/Emby/Jellyfish which then would then come with a set of Sanoid configuration lines. I'll open a new issue for that.

PR #98, opened 2019-06-05
Fix link to application ports.

martinfiers:
Small change to fix link to application ports.


davestephens:
Good spot, thank you :+1:

martinfiers:
Oh you're welcome, thanks for nice piece of work, I'm currently experimenting in a VM before I put it in production!

PR #101, opened 2019-06-12
adding wallabag

jesse-troy:
wallabag is a self-hostable PHP application allowing you to not miss any content anymore. Click, save and read it when you can. It extracts content so that you can read it when you have time.

I've been using this on my own server for a few months and it works great with the android app and the firefox addon.


davestephens:
Awesome contribution, thank you :+1:

PR #106, opened 2019-06-26
Vault secures credentials and separates them from config

seeReadCode:
Firstly, thanks to everyone who's worked on this lovely bit of software! I've enjoyed setting it up and tinkering with it.

I've been using vault to segregate credentials and config with the idea that I'm comfortable keeping my config in a private repo but rather not leave any credentials in there.  So, I figured I'd share this in case anyone else found this helpful.  Please let me know what you think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,12 +1,16 @@
-group_vars/all.yml
-group_vars/vpn_credentials.yml
 tests/test.yml
 site
-inventory
 .vagrant
 roles
 *.retry
 *.log
+.vault_pass
+
+
+# add ! for personal deploy branch
+group_vars/vault.yml  
+group_vars/all.yml
+
 
 # Vim
 [._]sw[a-p]
--- a/README.md
+++ b/README.md
@@ -99,6 +99,36 @@ download code from the internet and trust that it's going to work as you expect.
 8. Install the dependent roles: `ansible-galaxy install -r requirements.yml` (you might need sudo to install Ansible roles)
 9. Run the playbook - something like `ansible-playbook -i inventory nas.yml -b -K` should do you nicely.
 
+## Secure Repo Quick Start
+If you want to keep your credentials in your repo but secure them via [vault](https://docs.ansible.com/ansible/latest/user_guide/vault.html), my branch allows you to segregate critical secure info from configuration info and deploy without entering your sudo password everytime.
+
+To setup...
+
+```bash
+ansible-galaxy install -r requirements.yml
+cp group_vars/vault.yml.dist group_vars/vault.yml
+cp group_vars/all.yml.dist-vault group_vars/all.yml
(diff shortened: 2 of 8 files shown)


davestephens:
Thanks for this, I think it's a good idea :+1: However, I think the implementation needs some love to tidy it up a bit and make it work for everyone. Right now the changes are a bit too ingrained into the overall structure and setup of the playbook, IMO.

- The way the vault variables have been implemented will be difficult to work with once the vault is encrypted, such that you'll never be able to find them with a text editor. It's also not clear from the unencrypted files if the variables are in the vault or not. The generally accepted standard is to do something like this:

```
# random_application.yml
application_password: {{ vault_application_password }}

# vault.yml
vault_application_password: blahblah
```
This way, you can instantly see from the unencrypted code that the actual variable will be stored in the encrypted vault.

- I'm not clear why you've picked the variables you have to go into the vault, there are some random ones in there that don't seem vault-worthy?

- The changes you've made to `inventory` have made using the vault non-optional.

- The instructions included in `README.md` are way too verbose for the readme, IMO. Ideally I'd like to see a separate doc page for this stuff (as there isn't one you've included right now), and perhaps a note in the Quick Start section about how to encrypt variables, should you choose to. 

I'm currently working on some changes that will restructure the default variables and inventory somewhat (#89) - I'd like to work this into it but in such a way that a user can choose what they do/don't encrypt - not what we decide to for them. I'll get it pushed onto a branch soon and it'll start to make a bit more sense!

Thanks again for your work on this!

seeReadCode:
@davestephens thanks for the kind words and constructive feedback!  A few follow up notes and questions:

- I'm curious which vars would be considered to be unnecessarily vaulted.
- I meant to mention this in the PR rather than just the `README`, but the `inventory` change allows the user to deploy (and automate deploys) without entering the server's sudo password.  But yes, I see how tracking `inventory` might be a bridge too far.  N.b. one minor bug to be addressed if the `ansible_host` is to be pulled from vault: on deploy, the console currently displays `{ansible_host}`rather than the actual value.

TBH, I was suspicious of the vault/non-vault split as I implemented it here, but it seemed to be an expedient way to communicate the concept and hand off a workable implementation.  

A quick read through the config layout suggested on #89 makes it sound like it should be feasible to allow users to set vaulted vars without much change to the repo beyond documentation, `.gitignore` changes and including vault in the requirements.

Feel free to close this PR for now and I'll follow along with #89 to see how we can incorporate vault with that approach.  Thanks again!

davestephens:
I'm going to close this - with the structure of the project, all sensitive info is now contained within `inventories/your_inventory/nas.yml` and users can choose to vault/encrypt that should they want to. Doing so automatically is out of scope of what I'd like to achieve here.

Thanks for the contribution! 👍

PR #107, opened 2019-06-29
Port Management

animeai:
All container ports have been added as variables to all.yml.dist making port management much simpler. "ansible-lint" identifies no issues. The playbook runs without issue when building all containers. Some minor errors (port clashes) are also resolved with this commit.


davestephens:
Thank you for this, it's been on my list to do for ages! 

I'm going to hold off merging this until I've completed #89, but this will definitely be included when I have.

PR #111, opened 2019-07-23
Add Home Assistant

sammaes:
Adding Home Assistant instead of Hass.io.


davestephens:
Have merged, thanks! :+1: 

Had to tweak a few things:
- Both host networking and a port were set - can't have both so picked the port method. The Home Assistant docs are inconsistent on this one, so who knows which is best!
- Traefik config port was wrong
- Added time zone (important for stuff that relies on sunrise/sunset yada yada

PR #113, opened 2019-08-21
Make bitwarden_rs work with Traefik

koenvervloesem:
The bitwarden_rs container didn't work with Traefik. Removing the `traefik.backend` label does the trick, as it now creates the backend backend-bitwarden-hub for websockets and the backend backend-bitwarden-web for http. See https://github.com/dani-garcia/bitwarden_rs/issues/453


davestephens:
Odd, I'm *SURE* I tried this when I initially merged the PR! Anyway. thanks for the fix! :+1:

PR #114, opened 2019-08-24
update firefly configuration

al-sh-ac-san:
see https://github.com/firefly-iii/firefly-iii/issues/2379

### DIFF
--- a/tasks/firefly.yml
+++ b/tasks/firefly.yml
@@ -40,12 +40,12 @@
     ports:
       - "8066:80"
     env:
-      FF_APP_ENV: "local"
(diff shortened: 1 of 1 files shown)


davestephens:
Thanks for the PR! Has something changed in the image that requires the car name updates?

al-sh-ac-san:
Sorry, missed the question. What did You mean by car name? I see You merged it already, so You probably looked for Yourself that Firefly dropped 'FF_' prefix for env variables in 4.8.0.

davestephens:
**var** name - I got autocorrected!

PR #116, opened 2019-08-26
Timemachine to use SMB instead of AFP

ryanolf:
Apple has deprecated Time Machine over AFP in favor of SMB. This pull request would change ansible-nas to use a dedicated Samba server for Time Machine (via awlx/samba-timemachine) as well as the appropriate configuration for AVAHI to make the server discoverable.

I've been using this version of Time Machine with ansible NAS for 6 months and it works great.


animeai:
I would highly recommend keeping the original timemachine settings and having this as an option unless this is a direct drop in replacement. I don't have apple products so do not use this service so cannot test to see whether changes affect old backups or not. Shouldn't be too hard to duplicate the new playbook and have both with a note in the install instructions to only run one at a time with preference for samba for new installs.

davestephens:
Sorry for being so late to the party here. This one is kinda difficult for me to opine on - I don't have any Apple gear to test this, so I'm happy to rely on you guys to tell me it works :-) However, I would expect that when a user runs the playbook, the old container is stopped and the new one starts without manual intervention.

Also:
- What is avahi installed for, and is it required to be set up as a systemd service directly on the ansible-nas box?
- Please can you update the docs to reflect the image change (10/10 for using your own one btw! :1st_place_medal:)

ryanolf:
@davestephens Avahi is installed so that the Timemachine can discovered by the macs on the network via mDNS (even though it operates on a nonstandard SMB port).

Avahi can be used to advertise a number of different services, not just Timemachine. The present timemachine docker container includes Avahi in the container and operates on host networking. The issue with this is that if other containers also want to use avahi, they will also need to run Avahi on host networking, and I'm not sure they can do that.... all the avahi servers will be vying for the same port. In addition, if folks do install system avahi, I worry it might interact/conflict with the containerized avahi. Seems problematic to me, though I haven't tested it. I don't believe Avahi can be configured to operate on different ports in different containers like other services can. 

The solution in my container is to make sure avahi is installed and add the service to the list of those advertised. Other containers that need avahi can do the same without issue.

I will update the documentation per your suggestion.

davestephens:
OK, thanks for the explanation, I think you're right - in terms of it not being a pain in the ass to manage, and being reusable by other services, installing the system one seems like the best thing to do in this case.

Please can you add a section to the Time Machine docs to mention that Avahi will be installed *on the system* (ie not as a Docker container), and that config will be added so that Time Machine is discoverable? 

If you're feeling super-generous/motivated, perhaps you could merge #130 into your branch too? :smile:

ryanolf:
@davestephens In the latest commits:

-  I updated the timemachine.md document with the note about Avahi as well as with some additional info on the latest implementation relevant to both new users and those upgrading.
- I added a note to the upgrading.md file about a need to re-select the time machine disk. Since there are not official "versions" of Ansible-NAS I just refer to upgrading from "older" versions.
- the README.md file now refers to the new docker image and includes a link to the hosted upgrading.md document

Also, I believe that this change renders #130 irrelevant. The new samba-timemachine docker image runs only a single service and logs to STDOUT and thus logs are captured by Docker's built-in logging mechanisms and can be configured via Docker if the defaults aren't suitable. I noted this in the new timemachine.md docs.

davestephens:
Awesome! I rejigged the docs slightly - hope that's ok with you but meant I could get it merged today :+1:

PR #117, opened 2019-08-26
Adding homebridge support

ryanolf:
Added support for homebridge (https://github.com/nfarina/homebridge) via https://github.com/oznu/docker-homebridge, an Apple HomeKit bridge for connecting various smart things to Siri/HomeKit via plugins. Config is done by UI on a given port.


ryanolf:
I looked briefly at enabling traefik support for the homebridge UI, but traefik as it's set-up right now doesn't support proxying to containers that are on the host network (traefik itself is only on the docker bridge network).

ryanolf:
Revisited traefik, and it's easy to make it work, see https://github.com/davestephens/ansible-nas/pull/126. In case it ever does, added back the relevant connections.

davestephens:
Thanks for the contribution! :+1:  

I've tweaked it a little to remove to port config - this is going to be handled for all apps in #107. I've also removed the `| string` from every variable - it's a little cleaner to just ensure they are strings in the `all.yml` file. Oh, and I added docs. :smile:

ryanolf:
Great, thanks!

PR #118, opened 2019-08-26
Expose music folders to plex

ryanolf:
Plex has the ability to stream music and podcasts, as well as movies and TV. It's a minor thing but it might be nice to support this by default in Ansible-NAS.


davestephens:
A nice addition for those that still have mp3s ;-)

PR #119, opened 2019-08-26
Changed ansible-nas user to use "nologin" shell

ryanolf:
For security, in the event anyone gives the ansible-nas user a password (not sure why they would), we should probably set the shell to /usr/sbin/nologin. Probably not a big deal but doesn't hurt either. I did this so I thought I'd share.


davestephens:
Good spot, thanks :+1:

PR #123, opened 2019-08-27
Adding Jellyfin

jesse-troy:
Adding jellyfin media server. I've been testing this image on my server for several weeks now and it runs great.

There is an option to add GPU transcoding support, but I didn't include any of that. I might update the documentation with information on how to add it, but unfortunately I don't think we could really automate that since it will probably be different for everyone.

I used the format of the emby implementation, but in my testing I do not have network mode set to host. Not sure why the others are set this way (for upnp?) but I left it as it is.


davestephens:
Excellent addition, thank you! :+1:  Had to make a few tweaks to merge it, mostly issues of my own making by taking so long to get to this! :man_facepalming:

PR #126, opened 2019-08-29
traefik networking to host mode

ryanolf:
With traefik configured as it is now, it does not work with containers using "host" networking. This includes glances, emby, plex, tautilli, and others. Nonetheless, some of these include an optional "available_externally" flag and traefik labels... but they don't work. This pull request changes traefik to use host networking as well, which gives it access to all of the containers on the host and on the docker bridge network.


davestephens:
Nice spot, thank you! :+1:

PR #130, opened 2019-09-09
Separate config/logs and data directories for Time Machine container 

### LINKED ISSUE #105: Separate config/logs and data directories for Time Machine container
I keep my Docker home directory on the boot drive, an SSD with limited space. I'd like to avoid using it for large files such as backup data.

Other services tend to separate the config and data directories - it would be good if this was also true for the Time Machine service so that logs were kept under the Docker home directory and backups had the option of being stored elsewhere.

CherryKitten:
Split the directories for the timemachine container in two, making it
possible to save data on a different drive than logs.
Fixes #105


davestephens:
Thanks for the contribution! I'm sticking this one on hold for now pending #116

davestephens:
Closing as now covered by #116 and the new SMB based Time Machine :+1:

Issue #128, opened 2019-08-31
Option to use Traefik with self-signed wildcard certificate


koenvervloesem:
**Is your feature request related to a problem? Please describe.**

I need to configure Ansible-NAS to use Traefik with a self-signed wildcard certificate for local domain names instead of a Let's Encrypt certificate.

See **Additional context** for why I need this (access restricted to my local network) and how I implemented this.

**Describe the solution you'd like**

Currently, when you have `traefik_enabled: true` in `group_vars/all.yml`, Traefik is configured to use Let's Encrypt. I want to separate these two services so I can use Traefik with my own self-signed certificate.

For the sake of backwards compatibility, the default behaviour can stay like it is now: `traefik_enabled: true` enables Traefik and configures it with Let's Encrypt. If you want to use your own certificates, then that should be possible with extra options in the Traefik section in `group_vars/all.yml`, like:

```
###
### Traefik
###
traefik_docker_image: traefik:latest
traefik_data_directory: "{{ docker_home }}/traefik"
traefik_debug: "true"
traefik_tls_certificate: "{{ traefik_data_directory }}/cert.pem"
traefik_tls_key: "{{ traefik_data_directory }}/key.pem"
```

When these options are defined, the Traefik task in `tasks/traefik.yml` should install another `traefik.toml`than the current one, without the `[acme]` section and with the certificate and key files from `group_vars/all.yml` set up for the `https` entry point, like this:

```
  [entryPoints.https]
  address = ":443"
    [entryPoints.https.tls]
    [entryPoints.httpSSL.tls.defaultCertificate]
      certFile = "/etc/traefik/cert.pem"
      keyFile = "/etc/traefik/key.pem"
```

I can probably create a PR for this if you're OK with this approach. I can also document my setup (in **Additional context**) for anyone who wants to do the same.

**Describe alternatives you've considered**

I haven't considered any other alternatives. This solution looks like the most natural way.

**Additional context**

Traefik is awesome so I can use a subdomain for all my services, but I don't want to expose them to the internet. So I'd like to get the same experience that Traefik and Cloudflare DDNS give, but with access restricted to my local network.

The solution that I implemented recently and that has been working fine now for more than a few weeks, is the following:

- My local DNS server (Unbound) gives every host on my local network a `foobar.home` hostname, so e.g. `ansible-nas.home` for my Ubuntu Server machine running Ansible-NAS.
[...]

PR #132, opened 2019-09-17
Fix traefik v1.7

koenvervloesem:
The `traefik:latest` tag pulled in Traefik 2.0 for me today, which resulted in a failed Traefik configuration because it has breaking changes in the configuration file format. This PR fixes this by specifying `traefik:v1.7` for the docker image.

### DIFF
--- a/group_vars/all.yml.dist
+++ b/group_vars/all.yml.dist
@@ -274,7 +274,7 @@ samba_netbios_name: "{{ ansible_nas_hostname }}"
 ###
 ### Traefik
 ###
-traefik_docker_image: traefik:latest
+traefik_docker_image: traefik:v1.7
 traefik_data_directory: "{{ docker_home }}/traefik"
 traefik_debug: "false"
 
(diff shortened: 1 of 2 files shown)


davestephens:
I feel like by doing this we're fixing the wrong thing.

I've not looked, but how much work is involved in upgrading the config to
v2?

koenvervloesem:
Yes, I know that Traefik 2.0 would be the way to go eventually, and after a cursory glance it seems like the changes needed for Ansible-NAS are not that big, but Traefik 2.0 is still a release candidate (although for some reason they already switched the latest tag for their official Docker container to use 2.0) and poorly documented in comparison to the 1.7 branch.

I would definitely hardcode Traefik 1.7 for now, otherwise people's external access to Ansible-NAS services stops working when they update the Traefik docker container, like happened with me.

We/I can try working on a Traefik 2.0 migration when I have some more time. But I noticed there are still a couple of open issues and pull requests concerning Traefik (for instance [this one](https://github.com/davestephens/ansible-nas/issues/128), do you have an opinion about it, by the way?), so this rework will impact them.

davestephens:
OK, 2.0 is now GA - I'll merge this for now pending the upgrade!

davestephens:
Track upgrade in issue #142

PR #138, opened 2019-10-03
"Permission share dirirectories" – file: state: directory

technicalguy:
If you disable the bertvv.samba role (e.g. because you only want to use NFS, and not Samba), then general.yml fails in the "Permisison share directories" because they were not created. But this task specification can just require that they should be directories, and then it will make sure to create them if they do not exist, avoiding a "file/directory not found" error.


davestephens:
Nice workaround, thanks! :+1:

PR #139, opened 2019-10-03
add docker: syncthing/syncthing

technicalguy:
I'm not sure I got the labels for traefik correct...?

```
    labels:
      traefik.web.frontend.rule: "Host:syncthing.{{ ansible_nas_domain }}"
      traefik.enable: "{{ syncthing_available_externally }}"
      traefik.port: "8384"
```

### DIFF
--- a/docs/applications/syncthing.md
+++ b/docs/applications/syncthing.md
@@ -1,0 +1,35 @@
+# Syncthing: Open Source Continuous File Synchronisation
+
+Homepage: [https://syncthing.net/](https://syncthing.net/)  
+Github: [https://github.com/syncthing/syncthing](https://github.com/syncthing/syncthing)  
+Docker: [https://hub.docker.com/r/syncthing/syncthing](https://hub.docker.com/r/syncthing/syncthing)
+
+Syncthing is a continuous file synchronization program. It synchronizes files
+between two or more computers. It strives to fulfill the goals below in summary.
+Syncthing should be:
+
+1. Safe From Data Loss
+2. Secure Against Attackers
+3. Easy to Use
+4. Automatic
+5. Universally Available
+6. For Individuals
+7. For eveything else see the goals document
+
+## Usage
+
+Set `syncthing_enabled: true` in your `group_vars/all.yml` file.
+
+## Specific Configuration
+
+Open the web interface at :8384 to configure.
+
+The tasks/syncthing.yml sets
+```yaml
+    env:
+      STGUIADDRESS: ""
+```
+which disables admin GUI access from anywhere except localhost. You can either
+remove this (at which point it is visible to your whole network), or use
+Traefik. In either case you should consider setting an admin password in the
+syncthing GUI.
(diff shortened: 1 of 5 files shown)


technicalguy:
Maybe this should be linuxserver.io/syncthing docker image instead?

tasks/syncthing.yml
```diff
@@ -0,0 +1,22 @@
...
+    state: directory
+  with_items:
+    - "{{ syncthing_data_directory }}"
+
+- name: Syncthing Docker Container
+  docker_container:
+    name: syncthing
+    image: syncthing/syncthing:latest
+    pull: true
+    network_mode: host
```
davestephens: Please can you remove `network_mode: host` and replace it with the specific ports mentioned in https://github.com/syncthing/syncthing/blob/master/README-Docker.md

docs/applications/syncthing.md
```diff
@@ -0,0 +1,35 @@
...
+
+The tasks/syncthing.yml sets
+```yaml
+    env:
+      STGUIADDRESS: ""
+```
+which disables admin GUI access from anywhere except localhost. You can either
+remove this (at which point it is visible to your whole network), or use
+Traefik. In either case you should consider setting an admin password in the
+syncthing GUI.
```
davestephens: I'm not a fan of instructions that ask users to edit the `task/*.yml` files - it causes issues when users try to merge updates from master. Additionally, blocking GUI access to everywhere except localhost (once you've removed host-mode networking) will mean that it would need to be localhost from *within* the container...which is not exactly trivial :) My preference is this is removed and in the docs we tell users they need to set a password on first use (as is the case with many other applications)

tasks/syncthing.yml
```diff
@@ -0,0 +1,22 @@
...
+    volumes:
+      - "{{ syncthing_data_directory }}:/var/syncthing:rw"
+    env:
+      STGUIADDRESS: ""
```
davestephens: Should probably remove this per comments above

tasks/syncthing.yml
```diff
@@ -0,0 +1,22 @@
...
+    labels:
+      traefik.web.frontend.rule: "Host:syncthing.{{ ansible_nas_domain }}"
```
davestephens: This is correct but you also need to add this hostname to `traefik.toml`, see the file for examples of other applications.

davestephens:
Thanks for the contribution! I've added a few comments into your PR, also please can you add the application to the list in `README.md`. 

It may be worth trying out the linuxserver image - they solve annoying permissions issues by allowing you to specify a UID/GID that the container runs as. Ansible-NAS uses quite a few linuxserver images already if you need examples to check out :+1:

technicalguy:
Hi, thanks for the feedback, and sorry for not getting back sooner. I think it would be best to switch to the linuxserver image and also incorporate your other suggestions.

I haven't looked into this yet because I jumped ship to homelabos because of systemd integration (means I can make services depend on certain mounted drives).

PR #141, opened 2019-10-05
Add restic backup

technicalguy:


davestephens:
I've been playing about with this to try and come up with a way to make it more flexible - right now the only option is to back up to local storage (which could then be mounted elsewhere), which is pretty limited. Any thoughts on how backing up to s3 could be supported too?

Additionally, the ideal is that people shouldn't have to edit the individual app.yml files - the only changes required should be to your own `all.yml` - so I've pushed a change to your branch to support that.

Would appreciate your thoughts!

technicalguy:
Hi, sorry I've been offline for a long while.

I think [HomelabOS](https://gitlab.com/NickBusey/HomelabOS) has restic set up to use an S3 container (and you cannot backup to local storage). It is configured from within the main configuration file.

I guess it would be best to support both, configured from the main config file.

davestephens:
Stale + no demand, closing.

PR #156, opened 2019-10-16
Lidarr support

badcrc:
Lidarr works like Sonarr and Radarr but for music.

https://lidarr.audio/

### DIFF
--- a/group_vars/all.yml.dist
+++ b/group_vars/all.yml.dist
@@ -43,6 +43,7 @@ bazarr_enabled: false
 # Music
 airsonic_enabled: false
 mymediaforalexa_enabled: false
+lidarr_enabled: false
 
 # News
 miniflux_enabled: false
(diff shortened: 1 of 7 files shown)


tylergarcianet:
Nice to see this being included. I was able to use this to install Lidarr on my nas at home and then used your PR to model my after my PR to support Ombi.

group_vars/all.yml.dist
```diff
@@ -43,6 +43,7 @@ bazarr_enabled: false
 # Music
 airsonic_enabled: false
 mymediaforalexa_enabled: false
+lidarr_enabled: false
```
tylergarcianet: Might this fit better under `# Media Sourcing`?

tasks/lidarr.yml
```diff
@@ -0,0 +1,30 @@
+---
+- name: Create lidarr Directories
+  file:
+    path: "{{ item }}"
+    state: directory
+  with_items:
+    - "{{ lidarr_data_directory }}"
```
tylergarcianet: You shouldn't need `with_items:` if we're only creating a single directory.

davestephens:
Thanks for the contribution! (and your patience until I could find some time to merge this :+1:)

PR #161, opened 2019-11-02
Add photo support to Plex application

lmprice:
Add access to the photos path for the Plex application. This share
already exists, it just makes it available to Plex for photo views/uploads.


davestephens:
Cool addition, thanks! :+1:

PR #162, opened 2019-11-11
Add Step Certificates CA and ACME server

koenvervloesem:
Step Certificates is a private certificate authority (CA) and ACME server. With this pull request, Traefik can use it instead of Let's Encrypt (and other services on your network can use it too), so you are in full control of your certificate infrastructure.

This is a more robust and flexible approach than using a self-signed certificate such as in https://github.com/davestephens/ansible-nas/issues/128. I have been successfully running this branch here for a few weeks now on my Ansible-NAS box.

I don't use Let's Encrypt, so this needs some review, especially for the changes I made in `templates/traefik/traefik.toml` and `tasks/traefik.yml`. For instance, probably the latter task will fail now when you don't use Step Certificates because it refers to directories in the Step Certificates container directory. I don't know how to make this conditional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Ansible config and a bunch of Docker containers.
 * A backup tool - allows scheduled backups to Amazon S3, OneDrive, Dropbox etc
 * An IRC bouncer and web-based client
 * Source control with Gitea
-* SSL secured external access to applications via Traefik
+* SSL secured external access to applications via Traefik and Let's Encrypt or Step Certificates
 * A Docker host with Portainer for image and container management
 * An automatic dynamic DNS updater if you use Cloudflare to host your domain DNS
 * A Personal finance manager
@@ -66,6 +66,7 @@ Ansible config and a bunch of Docker containers.
 * [Radarr](https://radarr.video/) - for organising and downloading movies
 * [Sickchill](https://sickchill.github.io/) - for managing TV episodes
 * [Sonarr](https://sonarr.tv/) - for downloading and managing TV episodes
+* [Step Certificates](https://smallstep.com/certificates/) - A private certificate authority (CA) and ACME server for secure automated certificate management
 * [Tautulli](http://tautulli.com/) - Monitor Your Plex Media Server
 * [Telegraf](https://github.com/influxdata/telegraf) - Metrics collection agent
 * [The Lounge](https://thelounge.chat) - Web based always-on IRC client
--- a/docs/applications/step-ca.md
+++ b/docs/applications/step-ca.md
@@ -40,4 +40,4 @@ sudo step certificate install --all .step/certs/root_ca.crt
 ## Step Certificates as a CA and ACME server for Traefik
 If Step Certificates is enabled, Traefik will use it as a CA and ACME server instead of Let's Encrypt.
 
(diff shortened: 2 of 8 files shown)


koenvervloesem:
What's the best way to split this line in `tasks/step-ca.yml`:

`    command: "step ca init --name 'Ansible-NAS CA' --provisioner {{ ansible_nas_email }} --dns {{ ansible_nas_domain }} --address ':8443' --password-file secrets/intermediate_pass"`

on multiple lines?

koenvervloesem:
@davestephens If you have any questions about this pull request, feel free to ask them here or on Gitter.

koenvervloesem:
So, the long command is fixed now, split into multiple lines.

davestephens:
Looks like an interesting addition, and something I'd be happy to add once this PR has had a bit of spit and polish to tidy it up.

Few things stick out:

- The "set this flag, run the playbook, set the flag to something else, run the playbook again" doesn't sit well with me. If there's anything I've learnt from running this project, it's that people rarely read documentation before running it! Presumably StepCA will write out something once it's initialised, so you can use these file(s) for a check to stop the init being run more than once.

- Is there any documentation online around how to install the StepCA root cert on various operating systems? Presumably it's the same as installing any other root cert on an OS? I'd rather not have this in the Ansible-NAS documentation if possible. 

- The changes to Traefik seem to work on the assumption that StepCA will definitely be enabled, and/or has been run previously?

- The "run your own DNS to make this work" is...vague :-) (see documentation point above). How are you handling this part personally?

koenvervloesem:
> The changes to Traefik seem to work on the assumption that StepCA will definitely be enabled, and/or has been run previously?

Yes, this was why I was desperately trying to get your attention:

> I don't use Let's Encrypt, so this needs some review, especially for the changes I made in templates/traefik/traefik.toml and tasks/traefik.yml. For instance, probably the latter task will fail now when you don't use Step Certificates because it refers to directories in the Step Certificates container directory. I don't know how to make this conditional.

I'm not fluent enough in Ansible to do this. Can you give me a hint?

I'll work on your other remarks.

koenvervloesem:
> The "set this flag, run the playbook, set the flag to something else, run the playbook again" doesn't sit well with me. 

Fair point. The stepca task now checks for the existence of the ca.json configuration file. If it isn't found, it initializes the CA, and if it doesn't find the ACME provisioner in the configuration file, it creates the ACME provisioner. 

> Is there any documentation online around how to install the StepCA root cert on various operating systems? Presumably it's the same as installing any other root cert on an OS? I'd rather not have this in the Ansible-NAS documentation if possible.

There is [this](https://smallstep.com/docs/cli/certificate/install/), but it's only about the last command in the 'Add your root certificate to your clients' section in the documentation I added. The problem is that if you get one detail wrong in the steps I added under 'Add your root certificate to your clients', the whole setup doesn't work because the whole point is that you use a trusted CA. That's why I added these instructions in full. Otherwise you get users complaining about the setup not working. I would definitely leave this part in the documentation.

> The "run your own DNS to make this work" is...vague :-) (see documentation point above). How are you handling this part personally?

This really depends on everyone's setup, but I added an explanation of one possible way to do it, with OPNsense and Unbound DNS (the way I'm doing it).

So I think I have addressed all your points now, except for the one where I'm stuck and I hope you can help me with: when Step Certificates is disabled, Traefik should keep working with Let's Encrypt. Can you take a look at this?

koenvervloesem:
Any update on this, @davestephens?

davestephens:
Closing due to being developed against Traefik 1. Would be happy to revisit if someone wants to get this working with Traefik2 in an automated way that doesn't break the LetsEncrypt implementation.

PR #168, opened 2019-12-07
Doc fix - use `zpool` where `zfs` is incorrectly used

### LINKED ISSUE #167: ZFS Configuration doc uses `zfs` rather than `zpool` to set the `autoexpand` pool property
I'm new to ZFS so was pleased to find some relevant info in this project while having my first tinker with Ansible-NAS. However, the ZFS Configuration doc suggests turning `autoexpand` on for pools using:

```
sudo zfs set autoexpand=on tank
```

That gives me an error (on Ubuntu Server 18.0.4.3, but I doubt that's material) :

```
cannot set property for 'tank': invalid property 'autoexpand'
```

It seems `zpool` should be used instead of `zfs`:

```
sudo zpool set autoexpand=on tank
```

On the plus side, my ZFS knowledge has grown significantly digging into this 😄

codebrewer:
Fix #167 by enabling auto-expansion using `zpool` rather than `zfs`.

Briefly highlight that a pool has properties that apply to itself or to its filesystems.

### DIFF
--- a/docs/zfs/zfs_configuration.md
+++ b/docs/zfs/zfs_configuration.md
@@ -102,24 +102,26 @@ Our new pool is named `tank` and is mirrored. To see information about it, use
 `ashift`), use `sudo zpool destroy tank` and start over _now_ before it's too
 late.
 
-### Pool default parameters
+### Pool and filesystem properties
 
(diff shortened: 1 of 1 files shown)


davestephens:
Good spot, thanks! :+1:

Issue #172, opened 2019-12-27
Vagrant test script's ASCII art header renders incorrectly


codebrewer:
Not exactly critical! The launch script provided to spin up a testing VM prints some ASCII art but this doesn't render correctly, appearing as:

```
                     ._____.   .__
_____    ____   _____|__\_ |__ |  |   ____             ____ _____    ______
\__  \  /    \ /  ___/  || __ \|  | _/ __ \   ______  /    \__  \  /  ___/
 / __ \|   |  \___ \|  || \_\ \  |_\  ___/  /_____/ |   |  \/ __ \_\___ (____  /___|  /____  >__||___  /____/\___  >         |___|  (____  /____  >
     \/     \/     \/        \/          \/               \/     \/     \/
   a n s i b l e - n a s      https://github.com/davestephens/ansible-nas
```

instead of:

```
                     ._____.   .__
_____    ____   _____|__\_ |__ |  |   ____             ____ _____    ______
\__  \  /    \ /  ___/  || __ \|  | _/ __ \   ______  /    \\__  \  /  ___/
 / __ \|   |  \\___ \|  || \_\ \  |_\  ___/  /_____/ |   |  \/ __ \_\___ \
(____  /___|  /____  >__||___  /____/\___  >         |___|  (____  /____  >
     \/     \/     \/        \/          \/               \/     \/     \/
   a n s i b l e - n a s      https://github.com/davestephens/ansible-nas
```

codebrewer:
Addressed by #173

davestephens:
Fixed by #173 👍 (thanks!)

PR #173, opened 2019-12-27
Quote the Vagrant test script here document's delimiting identifier

codebrewer:
Quoting prevents interpretation of things like backslash characters.


codebrewer:
Addresses #172

davestephens:
Awesome, cheers! 💪

PR #176, opened 2020-01-05
add PyTivo to ansible-nas

bcurran3:
My first ever official PR and work with Ansible. I THINK I did everything right.

I have 3-4 other additions ready to go and a bunch of administrative PRs coming your way too...

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ Ansible config and a bunch of Docker containers.
 * [Plex](https://www.plex.tv/) - Plex Media Server
 * [Portainer](https://portainer.io/) - for managing Docker and running custom images
 * [pyLoad](https://pyload.net/) - A download manager with a friendly web-interface
+* [PyTivo](http://pytivo.org) - An HMO and GoBack server for TiVos.
 * [Radarr](https://radarr.video/) - for organising and downloading movies
 * [Sickchill](https://sickchill.github.io/) - for managing TV episodes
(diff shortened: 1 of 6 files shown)


bcurran3:
I wish I understood why GH thinks I deleted every line in the original code and then added the same code back as new! Maybe an encoding issue since I'm using Notepad++ on Windows to edit?

davestephens:
Hi @bcurran3, all of your commits are using CRLF line endings (which is why github shows the file as changed , which makes me think your git settings on your computer are a bit out of whack. 

Commits to git repos should be LF, with the user working in whatever line ending they choose (CRLF in your case if you *really* want to use Nodepad++). Git will help you with this - look at the autocrlf setting then fix your commits so we can see the actual changes.

I'd recommend taking a look at Visual Studio Code, it's significantly better and more modern than Notepad++.

animeai:
You can fix this by cloning your branch onto an ubuntu server or VM, then:
sudo apt install dos2unix

Then run in each directory you made edits:
dos2unix ./*

Copy the files back to windows via ssh and commit changes.

davestephens:
...or just set up git correctly in the first place!

bcurran3:
I can save in Unix format easily in Notepad++ now that I realize the problem.

Closing and opening again new.

davestephens:
Look at the `core.autocrlf` setting, I think you have that set wrong. That way you don't need to remember to edit in specific formats, git sorts it out for you.

PR #184, opened 2020-01-05
Samba advanced options added for Windows Server environments

bcurran3:
Though these values already default to true from https://github.com/bertvv/ansible-role-samba this makes it easier to set them to false for users with PDCs.


davestephens:
This is a nice change, thank you. I will merge after #193 :+1:

PR #186, opened 2020-01-06
Make docker images variables

animeai:
Allow for simple testing of new versions, change to a different image if you have a favorite or make sure you stay with an older version of a container. Ideally, once this set of PR are finished, the /tasks/*.yml should never need to be edited unless something breaks, all the configuration settings should be editable as variables.


davestephens:
The shell script permissions have been changed, which has broken the build.

animeai:
I actually have no idea how or why the shell scripts permissions got edited... I didn't modify those files to my knowledge!

animeai:

davestephens:
FYI you don't need to keep merging latest changes into your PRs, it makes reviewing them very hard as you end up including stuff for me to review that I've merged a while back :confused: 

Aside from that, I've been thinking about this one a bit, and I'm not sure current me agrees with me from a month ago. I'm curious as to the benefit of having images in variables. 

- *For the most part* there is no reason to ever change an image being used. Most are set to latest anyway, except for things like some mysqls, Traefik. Development purposes yes, but it doesn't make sense to put everything into variables just for that, because it would be submitted as a PR anyway.

- Looking up what image is used for an application has an extra two levels of obfuscation (task.yml -> all.yml -> nas.yml)

- Supportability - I can already see people raising issues about applications that don't work, only to find they've pinned some ancient image or using some new/different one to everyone else

Is there something I'm missing here? Do you have real situations where you have a "favourite" image over one picked in the project, but for some reason haven't submitted as a PR?

animeai:
Hmmm. I see where you're coming from. I will admit I was thinking more from the development point of view than the typical user. If this is the case and you don't think there is value in this or another PR then feel free to close them. If you want or need one or more if the other PR remade without the upstream merges let me know.

Also,  noted in the upstream merges. I didnt realise this was an issue.

davestephens:
OK, I'm going to close this one out. We can always revisit on a case by case basis (ie like with Traefik), but right now I don't think there's value in making things configurable just for development purposes.

PR #187, opened 2020-01-06
Make transmission and transmission openvpn work together

animeai:
Personally I use normal transmission for private trackers and vpn transmission for the rare occasion I need to use a public tracker so having both working together with separate watch directories etc makes sense to me.


animeai:

davestephens:
As far as I can see, this has been covered by other PRs and recent updates - feel free to reopen if I'm missing something! 👍

PR #188, opened 2020-01-06
Map ports to all.yml

animeai:
My previous PR that mapped ports to variables was so long ago, over half the current containers didn't exist.

As part of this PR, I've fixed the port conflict between thelounge and another container (I forget which!) and fixed a lot of errors on the traefik port. When specifying the traefik port, you should always use the mapped port, not the internal container port! Using variables of course prevents this in the future


davestephens:
You've changed the permissions on the shell scripts, which is breaking the build....

animeai:
@davestephens tested in VM with no errors - should be good to go!

davestephens:
Please can you cover #216 with this too? It will allow people to run AdGuard/PiHole internally should they choose to, as well as make us slightly less opinionated about what should run where...

animeai:
@davestephens I'll take a look at that when I get a chance. It shouldn't be too difficult I hope!

animeai:
@davestephens traefik now has configurable ports :) 

This has been tested in VM and should be ready to merge. Since the most recent commits, I did not retest all containers, only traefik. The traefik UI is accessible on the selected port and traefik.toml is written correctly.

PR #193, opened 2020-01-06
Rework Ansible-NAS Configuration

davestephens:
This PR totally reworks how Ansible-NAS configuration is structured, which will make updates simpler and first-time configuration easier.

- `group_vars/all.yml` is tracked in the repo and provided as a default configuration.
- Users create an inventory directory to store their own config by copying the provided sample.
- User customisations are added to `inventories/<my_inventory>/group_vars/all.yml`, which by ansible variable precedence, will override `group_vars/all.yml`.

Then when a user wants to pull the latest from master:

- They pull the latest from master
- Enable any new pieces of config they're interested in by editing their `inventories/<my_inventory>/group_vars/all.yml`.

The intention is that once this and the port mapping stuff is done, v1.0 will be a thing and we can work out how to create new version from thereon if people like that idea. Or we can just keep telling people to update from `master`, I'm easy.

Issue #109, opened 2019-07-17
Allow configurable subdomains


animeai:
ansible-nas/templates/traefik/traefik.toml lines 181 onwards.

Currently traefik will be requesting certificates for all subdomains whether we require them or not. This is not good practice as we will be hitting the letsencrypt daily limits when initially setting up. It would be better to use an ansible task to create the traefik.toml file based upon the services loaded. This is a little beyond my ansible expertise though! Currently I can only think that a series of plays would be required to do it building onto a variable string if a task is enabled before using a sed command to merge into the traefik.toml file... Something like: (shameless copy/paste of code form elsewhere to save typing!)

```
        - name: Initialize an empty list for our strings
	  set_fact:
	    my_strings: []

	- name: Setup a string variable
	  set_fact:
	    my_name: "Max"
	 
	- name: Append string to list

	  set_fact:
	    my_strings: "{{ my_strings + [ my_name ] }}"

	- debug: var=my_strings

	- name: Append another item to the list
	  set_fact:
	    my_strings: "{{ my_strings + [ 'Power' ] }}"

	- debug: var=my_strings
```

davestephens:
I'm not sure your assumptions are correct. Traefik only requests *one* certificate, and all of the various subdomains are requested as [SANs](https://knowledge.digicert.com/solution/SO9440.html) - so there's no cost in having many subdomains.

That said, there is an open PR (#44) to initially request the LetsEncrypt cert as a staging cert rather than production - it just needs some work before I'm happy to merge it.

animeai:
Ahhh, I've misread the code then!

Looking at #44 I can probably use that to make the subdomains editable through variables and only request the active containers too. I was unaware that I could pass variables from ansible straight into the traefik.toml like that. Shows how much I have to learn about the software still!

davestephens:
I'm not sure it needs changing from how it currently works, unless I'm
missing something?

Right now one cert is requested when Traefik starts - if you enable/disable
containers no new cert is required.

With what you're proposing a new cert would be requested every time you
enable/disable something, which could happen many times over a short period
of time - ending in you getting rate limited.

The only time a new cert is requested is if the existing cert is close to
expiry, or if a PR is merged with a new container that adds a new hostname
to the list of SANs.

animeai:
[...]

PR #201, opened 2020-01-07
Make subdomains editable

animeai:
I've needed to tweak subdomains to make them a little more memorable and I don't think we should be editing the /tasks/*.yml files as these get overwritten.


davestephens:
In theory I like this change, but I think it needs some documentation per the discussion in #109.

People need to be aware that changing subdomains and restarting traffic counts towards the LetsEncrypt rate limit. I *believe* that changing the cert SANS (subdomains) will count towards the "duplicate certs per week" limit imposed by LetsEncrypt, which is 5. e.g., you can play with the subdomains and re-request a certificate five times in one week before you get blocked and have to wait.

There is a way around this which would be to use the staging LetsEncrypt environment, but #35 was never completed which then adds a manual step in between switching between staging/prod.

animeai:
@davestephens Sounds good to me, I'll word something up after the #193 merge. It'll be a headache if I go changing install instructions now!

animeai:
Looking at the rate limits https://letsencrypt.org/docs/rate-limits/ - "Certificates registered per domain" is capped at 50 per week. The Duplicate Certificate rate is 5 per week though - I think we're more likely to be hitting this one.

We will also have to be careful in the future if we hit over 100 subdomains as 100 is the maximum names per certificate.

bcurran3:
> I think we're more likely to be hitting this one.
> 
> We will also have to be careful in the future if we hit over 100 subdomains as 100 is the maximum names per certificate.

I think this is all VERY conditional depending on a user's configuration; very much a documentation warning... not so much a we-need-to-be-careful-with-the-setup-scripts-situation. Ansible-NAS could have the possibility to setup 20,000 sub-domains but if a user's configuration only consists of 20... not a concern.

Something along the lines of:

WARNING: When using LetsEncrypt blahblahblah, use of more then X sub-domains may cause blahblahbalh. Please refer to https://linkblahblbahblah

image

davestephens:
@bcurran3 If you look at the Traefik config you'll see that we request a cert for *all* subdomains whether they are enabled or not, so @animeai's concerns are valid. 

We do this so the certificate is only re-requested when new applications are added, but it does come with the downside of requesting a lot of SANs per cert - which as stated above, has a limit of 100.

bcurran3:
Thanks. Good to know. One of these days I need to checkout Traefik but don't think I personally have a need since I use a VPN to home all the time.

davestephens:
Closing as now supported via wildcard certs and shift to ansible roles 👍

PR #203, opened 2020-01-07
Make docker options in tasks variables

animeai:
As we are likely moving to a version of ansible-nas that should be much more updatable without needing to merge changes to the all.yml file manually, we should be in a position where we never need to edit the /tasks/*.yml files. I've needed to tweak some of the memory limit settings on the containers and being able to tweak the restart policies is very useful when you have a misbehaving container.


animeai:

davestephens:
OK, finally got to checking this out.

Memory I can understand wanting to be configurable, but why the restart policy? Is there ever a situation where you *wouldn't* want Docker to keep applications alive for you?

animeai:
@davestephens Mostly this is for misbehaving containers. I want them to crash and stop so I can see the logs at the point of crash :) I don't want to have to manually modify things. Some may also want to change a particular container to "always" instead of "Unless stopped". That was my reasoning.

PR #208, opened 2020-01-09
Update test-vagrant.sh

joshmello:
- Test script needs to copy all.yml not all.yml.dist since it was renamed.
- Also added a -p flag for vagrant to provision after any changes instead of destroying and rebuilding.
- Basic formatting from Prettier.


davestephens:
Awesome addition, thank you :+1:

PR #214, opened 2020-01-12
Logitech Media Server addition

bcurran3:

Fixes #

**Any other useful info**:

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -42,6 +42,7 @@ lidarr_enabled: false
 # Music
 airsonic_enabled: false
 mymediaforalexa_enabled: false
+logitechmediaserver_enabled: false
 
 # News
 miniflux_enabled: false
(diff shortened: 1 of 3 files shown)


davestephens:
Hi, thanks for your contribution. 


Few things missing:
- docs
- ports listed in the relevant doc file
- the hostname is missing from traefik.toml
- add the application to the readme

Also:
- please remove the "last modified by comments", this is unnecessary when files are version controlled
- does logitech media server modify files it serves? If not, then we don't need the rw permission variable in `all.yml` and just fix the permission to `ro`
- The user and group id variables in `all.yml` aren't used.
- the mediaserver webport variable isn't used,
- the port isn't specified to Traefik so external access won't work as it's configured due to multiple ports being mapped. See other applications for examples.

bcurran3:
This was a mistake, was supposed to go to my fork only.
Will submit here later.

Issue #217, opened 2020-01-13
Homeassistant with host networking


andrewboie:
**Is your feature request related to a problem? Please describe.**
I've set up Homeassistant on my ansible-nas. It appears to be set up in "bridge" mode with its own private network. However, I can't get it to detect my LIFX bulbs on my LAN, and from some googling around it seems you need the container running in "host" mode for it to find them.

**Describe the solution you'd like**
A configuration option to create the Homeassistant container with host networking.

**Describe alternatives you've considered**
- Change the default to host networking for Homeassistant.
- A documentation note on how to configure this manually.

davestephens:
Thanks for the big report. Have you tested flipping the container to host-mode networking to see if it fixes it?

Adding `network_mode: "host" and removing `ports` from the homeassistant task yaml should do the trick.

andrewboie:
> Adding `network_mode: "host"` and removing `ports` from the homeassistant task yaml should do the trick.

Yup, that worked for me. Homeassistant was then able to detect my LIFX bulbs. It also found my HiFi stereo and the Harmony Hub. I think the "discovery" feature requires host networking.

davestephens:
Excellent. If you want to raise a PR I can get it merged ASAP 👍

PR #218, opened 2020-01-13
homeassistant: use host networking

andrewboie:
Homeassistant discovery requires that the docker container be
configured for host networking.

See: https://www.home-assistant.io/docs/installation/docker/

Fixes: #217


davestephens:
👍

PR #220, opened 2020-01-14
Vagrant setup uses inventories/group_vars

Vinibo:
Vagrant tests are now using the same file structure as regular ansible targets. This way, vagrant tests can act as documentation. It also simplify customized deployements in Vangrant by avoiding modifying all.yml.


davestephens:
I recently adjusted how the playbook testing worked, so have used your technique with the already-existing testing inventory. Thanks for the pointers 👍

axelsimon:
I think the `ansible.inventory_path = "tests/inventories/integration_testing/inventory"` line was placed in the wrong section.
I've fixed that in another change to the Vagrantfile for which i've made a PR (#451).

PR #221, opened 2020-01-16
Add Avahi for local-network .local resolution

Vinibo:
Add/Enable Avahi daemon in order to broadcast hostname for mDNS resolution.
mDNS-compatible computers will be able to resolve the NAS address through `<hostname>.local`.

It also enable autodiscovery of Samba shares by other computers on the same network.

**Any other useful info**:
More information on ArchLinux's wiki: https://wiki.archlinux.org/index.php/Avahi


davestephens:
Cool addition!

BUT! :smile: 

We already have Avahi installed by the time machine task: https://github.com/davestephens/ansible-nas/blob/master/tasks/timemachine.yml

I think we need to do something like this:

- Remove Avahi from the timemachine yaml
- Add a check to timemachine.yml to see if the Avahi package is installed, if not then error. If yes then carry on and install the timemachine Avahi config

Would appreciate your thoughts!

Vinibo:
Thanks for the suggestion, I was not aware of this dependency.

I've made the changes, and I also added more doc (Avahi and modified Timemachine doc).

PR #223, opened 2020-01-20
Installation pymodule update

Malpractis:
Existing Ansible installation instructions will result in version 2.5.2 rather than the current 2.9.x version.
Official Ansible documentation has been updated as per the change in docs/installation.md.

Additionally the docker-py module is deprecated and has been replaced with the 'docker' module.

Fixes #

**Any other useful info**:
These two changes are needed for the upcoming Traefik 2.1.1 PR I will be submitting once finished testing.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -11,6 +11,12 @@ You can run Ansible-NAS from the computer you plan to use for your NAS, or from 
 
 2. Install Ansible:
 
+    `sudo apt update`
+
+    `sudo apt install software-properties-common`
+
+    `sudo apt-add-repository --yes --update ppa:ansible/ansible`
+
     `sudo apt install ansible`
 
 3. Clone Ansible-NAS:
--- a/tasks/docker.yml
+++ b/tasks/docker.yml
@@ -5,7 +5,7 @@
     state: present
   register: result
   until: result is succeeded
-  
+
 - name: 'Check if docker-py is installed'
   shell: pip3 show docker-py
   failed_when: docker_py_check.rc != 1 and docker_py_check.rc != 0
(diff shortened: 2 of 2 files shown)


davestephens:
Funnily enough I had the docker change on a branch on my laptop, I just hadn't tested and pushed it yet 😁

Thanks for the ansible addition too - I want to run the playbook through before merging this as Ansible have a habit of adding lots of new deprecations with every release.

Malpractis:
Haha nice!

Yeah no worries. I've run it though a heap of times on ubuntu 18.04 with the 1000 other changes for traefik 2 so I don't expect any issues. :)

davestephens:
Agreed, I just want to check the case where both docker and docker-py are
installed. We should probably remove docker-py in that situation...

Malpractis:
Ah yeah. I do actually have a task to remove docker-py I can chuck in if you want? They will not run happily together, I found that out the hard way hah

davestephens:
Yes please, otherwise anyone that runs this over the top of an old version
will get b0rked.

Malpractis:
There we go, added a check for docker-py and if present remove docker-py.

The check will throw an error if docker-py is not present, hence why I've set it to ignore errors. I'm not sure of a nicer way to handle it though. You got any ideas?

Malpractis:
Got something in there to fix the failed docker-py check on subsequent runs without docker-py being installed. 

Not sure why the check is failing now.

davestephens:
It would be cleaner to use the `pip` module with `state: absent` which would save having to do a check then remove based on the check...

Malpractis:
Ah yeah, I wasn't sure if it would be good practice to have it always run the docker-py remove like that. But I guess it doesn't do anything once it's gone.

If you're happy with that I'm happy with that hah

animeai:
Very nice :) looking forward to testing this!

davestephens:
Nice :+1:

PR #227, opened 2020-02-01
Migrate Containers to Docker Network from host

Malpractis:
With Ansible & docker-py(now docker) updated, we are able to create a custom docker_network with ansible.

This PR migrates all containers to use a custom docker bridge network.

**Any other useful info**:
I've tested most containers myself, and will continue to do some more testing in the coming days. 
The main ones I can't adequately test are the home automation ones as I don't have any devices to link to them.


davestephens:
Looks interesting!

I think I'm missing something though, what's the benefit of making this change?

Malpractis:
Main benefit of a custom docker bridge network over using the default bridge is that it will allow containers to communicate with one another using hostnames rather than only being able to communicate via IP address (note I haven't actually tested this, I'm just going off of the [official docker documentation](https://docs.docker.com/network/network-tutorial-standalone/)). Additionally the official docker documentation recommends against using the default bridge in production. 

A practical example would be for something like Jackett in radarr/sonarr, if containers can use hostnames, you can configure the URL to something like http://jackett:9117. Which will also prevent silly things happening like users using the external URL (the traefik one) resulting in traffic going outside the LAN then back in for internal only traffic.

davestephens:
I believe (it's not actually documented anywhere) the reason that Docker recommend *not* using the default bridge and creating user-defined networks per application is to enforce segregation. Obviously all applications on the default bridge network can talk to each other, and furthermore linked applications share environment variables too - and so if you didn't understand this there could be security concerns.

In our situation, we don't care, and it's actually something we want (IE your example above). 

That said there is value in being able to resolve containers by hostname, but I don't think there's value in letting users change the name of the network - at least, not at this point in time with the applications we have. 

I also don't think there's any point in letting users change the `hostname` (should be the same as `name`). `network_cli_compatible` can be dropped as it'll be the default in Ansible 2.12, and before that I can't see a situation where it'd be useful?

Also I see you've included a change to Plex - please include that and any other unrelated changes in a separate PR.

Malpractis:
There we go.

I've removed the networks_cli_compatible, I can't 100% remember why I put it in there to be honest. Might have been looking at old doco and forgot to remove it from my private branch. 

I've made the hostnames static.

Reverted the migration of some containers to use docker network instead of host network. 

I've left the docker_network_name simply because it's only one variable, which is used in 114 files. Personally I have a naming convention for my physical network that I carry over into my virtual networks, so that's why I've included it.

Malpractis:
Just remembered why I had networks_cli_compatible configured:
`TASK [Traefik Docker Container (HTTP Challenge)] ***********************************************************************************************************************************************************************************************************************************
[DEPRECATION WARNING]: Please note that docker_container handles networks slightly different than docker CLI. If you specify networks, the default network will still be attached as the first network. (You can specify purge_networks to remove all networks not explicitly
listed.) This behavior will change in Ansible 2.12. You can change the behavior now by setting the new `networks_cli_compatible` option to `yes`, and remove this warning by setting it to `no`. This feature will be removed in version 2.12. Deprecation warnings can be disabled
 by setting deprecation_warnings=False in ansible.cfg.`

That's seen on ansible version:
`ansible@ansible-nas:~/ansible-nas-merged$ ansible --version
ansible 2.9.4
  config file = /home/ansible/ansible-nas-merged/ansible.cfg
  configured module search path = [u'/home/ansible/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.17 (default, Nov  7 2019, 10:07:09) [GCC 7.4.0]`

PR #228, opened 2020-02-02
WIP: Add Syncthing service

Vinibo:
This PR adds the Syncthing Service, a peer-to-peer synchronization service for files.
The doc also contains a Quick start in order to be able to setup a simple 'network'.

**Any other useful info**:
Like other services in ansible-nas, the Syncthing service uses the UID/GID 0, but this triggers a warning in the web UI regarding running as a privileged user. Might be tempted to create a user for Syncthing.

This PR does not auto-configure shares such as Documents, Music,... I wanted to do it at first but since everything is in a single XML file, including generated IDs, I opted for a guide instead.


Vinibo:
I tried implementing user/group for Syncthing but the new UID/GID cannot write in the samba shares (`/mnt/Volume3/music` for instance).

A way I found to mitigate this is by making the Syncthing user a member of `ansible-nas` group. In its current state, I cannot refer the `ansible-nas` GID as we only create it with a name and let the system decide its GID (most of the time, it would be 1001 but we never know).

The pull request #232 sets both UID/GID in order to be referenced here.

davestephens:
Closing - covered in #139 .

PR #232, opened 2020-02-05
Feature/predictable ansible uid gid

Vinibo:
Lets configure UID/GID for the user/group ansible-nas. It lets other applications be a user of the group ansible-nas, giving read/write access to the samba share volume without clashing with other applications.

**Any other useful info**:
I was trying to add user/group for Syncthing #228 and hit a problem when Syncthing was trying to write to `/mnt/Volume3/music/` getting permission denied. This is an attempt at mitigating the issue.


Malpractis:
That's great, nice one! I was thinking that we needed this just a few days ago.

davestephens:
Firstly, sorry for the delay getting back to you. I think this is a nice (and useful) change.

Secondly - I think this requires a little more thought before I'm happy to merge it.

This solution works nicely for new users, but if an existing user were to pull the playbook and run it, they could potentially end up changing their ansible-nas user id to 1002 (e.g. mine is 999), and thus breaking the perms on everything that's currently permissioned with the `ansible-nas` user (same applies to the group). For people not so familiar with Linux (there are plenty that use Ansible-NAS), this could land them with all kinds of mess to fix.

Therefore I think it requires an automated check in `user.yml` that checks to see if the user/group already exists. If it does, check the uid/gid against `ansible_nas_uid` and `ansible_nas_gid` and check the IDs match before carrying on. If they don't, the playbook should exit with an error.

What are your thoughts?

PR #236, opened 2020-02-08
Account for new Sanoid dependency in docs.

DLLarson:
Sanoid now requires an additional perl package that wasn't installed by default on Ubuntu 18.04 server.

-Dale


davestephens:
Thanks for the doc update :+1:

PR #238, opened 2020-02-20
Add Hddtemp monitoring to Telegraf/InfluxDB/Grafana.

ckoehler:

Add temperature monitoring to InfluxDB through Telegraf, for use in Grafana.

N/A

**Any other useful info**:


ckoehler:
This actually doesn't persist through reboots, because the IP address may change. So this may be related to #227 to be able to set static IP.

davestephens:
Thanks for the contribution! 

Just wondering why you didn't add hddtemp to `ansible_nas_extra_packages` and configure telegraf accordingly? Seems overkill to use a container for a single apt package, especially when you have to mount all of `/dev` to make it work.

The IP address of Ansible-NAS should be fixed, I'll add something to the docs, so the method used elsewhere in `telegraf.conf` to get the IP will work nicely for you.

ckoehler:
Good point, I didn't realize the containers could reach the host. I'll do it the other way.

PR #240, opened 2020-02-29
Added config variable to override nextcloud users data directory

codemeister64:

This PR adds the ability to specify a custom users data directory path for Nextcloud.
According to [docs](https://hub.docker.com/_/nextcloud/) we can mount volume to default data path "/var/www/html/data", which is used to store users data.

*My personal use case:*

I have 2 SSD (one used for OS, boot, and swap, the second one - for home directory) and several HDD.
All my docker containers, configs, etc, use the second SSD, but I want to store user's files on my HDD. In my opinion data storage should be split from the application (nextcloud) filesystem. That is the aim of this PR.

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -590,6 +590,7 @@ glances_port_two: "61209"
 ###
 nextcloud_available_externally: "false"
 nextcloud_data_directory: "{{ docker_home }}/nextcloud"
+nextcloud_users_files_directory: "{{ docker_home }}/nextcloud/nextcloud/data"
 nextcloud_port: "8080"
 
 ###
--- a/tasks/nextcloud.yml
+++ b/tasks/nextcloud.yml
@@ -6,6 +6,7 @@
   with_items:
     - "{{ nextcloud_data_directory }}/nextcloud"
     - "{{ nextcloud_data_directory }}/mysql"
+    - "{{ nextcloud_users_files_directory }}"
 
 - name: Nextcloud Mysql Docker Container
   docker_container:
@@ -31,6 +32,7 @@
         - nextcloud-mysql:mysql
     volumes:
       - "{{ nextcloud_data_directory }}/nextcloud:/var/www/html:rw"
+      - "{{ nextcloud_users_files_directory }}:/var/www/html/data:rw"
     ports:
       - "{{ nextcloud_port }}:80"
     env:


davestephens:
I like the idea, but the implementation feels a little [Inception](https://www.imdb.com/title/tt1375666/) - you're mounting a mount within another mount, which is asking for trouble.

Rather than create an extra mount into Nextcloud, just separate Nextcloud and Mysql. Something like this might be better:

```
nextcloud_data_directory: "{{ docker_home }}/nextcloud/nextcloud"
nextcloud_db_directory: "{{ docker_home }}/nextcloud/mysql"
```

...then adjust MySql accordingly.

codemeister64:
> I like the idea, but the implementation feels a little [Inception](https://www.imdb.com/title/tt1375666/) - you're mounting a mount within another mount, which is asking for trouble.
> 
> Rather than create an extra mount into Nextcloud, just separate Nextcloud and Mysql. Something like this might be better:
> 
> ```
> nextcloud_data_directory: "{{ docker_home }}/nextcloud/nextcloud"
> nextcloud_db_directory: "{{ docker_home }}/nextcloud/mysql"
> ```
> 
> ...then adjust MySql accordingly.

Yeah, I agree that we should separate MySQL too. But in this case, Nextcloud still will store the user's files inside its PHP application folder, which is the whole Nextcloud mount ({{nextcloud_data_directory}}) with PHP code, configs, etc. I think it is better to separate user's files from Nextcloud application. According to docs, there is another option - env variable ```NEXTCLOUD_DATA_DIR (default: /var/www/html/data. Configures the data directory where nextcloud stores all files from the users.)```,  I can try to use it.

So, according to this concept, we will have three variables
 
```
nextcloud_root_directory: "{{ docker_home }}/nextcloud" 
nextcloud_data_directory: "{{ downloads_root }}/nextcloud/data"
nextcloud_db_directory: "{{ downloads_root }}/nextcloud/mysql"

// nextcloud_root_directory: mounts from docker_home. Gives access to Nextcloud configs and has good performance due to the use of SSD (in my case)
// nextcloud_data_directory: mounts from any of my data folders, probably from ZFS data-sets (in my case it's one of the HDD`s). It should be configured by NEXTCLOUD_DATA_DIR env variable.
// downloads_root here is just an example to better explain the idea, this folder may be not shared via samba/NFS or to other applications (such as transmission)
// nextcloud_db_directory: mounts from any of my folder where I want to store Nextcloud DB.
```

What do you think about it?

PR #249, opened 2020-04-16
Use hostname defined in all.yml for main hostname in traefik template

bcurran3:

Change to use the variable defined hostname instead of hardcoded "ansible-nas" 

Fixes # n/a

**Any other useful info**:
None, simple and concise.


bcurran3:
Please help me understand the failure. Something about the trailing space removed? Or testing problem?

davestephens:
Nice change, thanks!

> Is this a "me thing?"

It was caused by an update to `mkdocs-material` - the theme I use for the docs pages. I've fixed it on master so happy to merge this in :-)

PR #252, opened 2020-04-16
Add advanced Samba options to all.yml for Windows Server environments

bcurran3:

REF: https://github.com/davestephens/ansible-nas/pull/184

Though these values already default to true from https://github.com/bertvv/ansible-role-samba this makes it easier to set them to false for users with PDCs.

Fixes # n/a

**Any other useful info**:
I have Windows servers too!

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -157,6 +157,9 @@ samba_domain_master: true
 samba_local_master: true
 samba_preferred_master: true
 samba_wins_support: true
+#samba_workgroup: "WORKGROUP"
+#samba_realm: "COMPANY.local"
+
 
 # Where stuff downloaded will be stored
 downloads_root: "{{ samba_shares_root }}/downloads"
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -149,18 +149,6 @@ docker_storage_driver: overlay2
 # This path will be mounted to backup containers, Duplicati
 samba_shares_root: /mnt/Volume3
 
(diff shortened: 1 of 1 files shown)


davestephens:
I'm not keen on having a bunch of options 99% of people will never change in the config - especially above the shares section - which 99% of people will change :-) Also, the link provided is to the root of the Samba docs, there is a ton of stuff there that *isn't* configurable through Ansible-NAS.

I think it would be better to link to the [`bertvv.samba` Ansible role Github page](https://github.com/bertvv/ansible-role-samba), along with a note of something like "Further Samba configuration is possible through Ansible-NAS, see <blah> for more info". This would probably be best placed at the bottom of the Samba section too.

It may not be immediately obvious (especially now it has grown so big), but `group_vars/all.yml` is roughly ordered in the likelihood that people will need to change stuff...

bcurran3:
So I should move it towards the bottom of all.yml?  No problem, I can do that. Personally I think best to keep all related items grouped together. Let me know if you change your mind.

Will change link.

Prepare for my "Return rage: Rage against the NAS"

davestephens:
No, towards the bottom of the samba section. But only a link and a note,
not all the options - because most won't use them (as you say).

bcurran3:
Changes made. I'm hoping they are amicable to all.

davestephens:
With the move to roles and nas.yml, users can add these settings straight into their config if they want to, there's no need for placeholders in all.yml any more.

Issue #253, opened 2020-04-16
Make memory: in all tasks a variable in all.yml


bcurran3:
**Describe the solution you'd like**

Convert memory: statements in tasks to use a variable defined in all.yml. Why? Why not, everything else is. No, really I like to give some containers some more breathing room as my server has plenty of RAM. Plex's pseudo recommendation is 2g even though it works in 1g. I like to give it 8g. (Yeah, overkill.) I give more RAM to Transmission too for better caching.

**Describe alternatives you've considered**

Continue to manually update the task files myself.

**Additional context**

None.

I'm willing to do the PR, but it would be awhile. I'm many many many commits behind and have other container apps I want to submit first.

PurpleNinja225:
I think a great way to acomplish this would be to edit the `roles/ansible-nas-docker/templates/daemon.json`  to look a little somthing like this:

```json
{
  "data-root": "{{ docker_image_directory }}",
  "storage-driver": "{{ docker_storage_driver }}",
  "memory": "{{ docker_default_memory }}"
}
````
add the variable to roles/example_app/defaults/main.yml
```yml
example_app_memory: "{{ docker_default_memory }}" 

and update the ../tasks/main.yml

```yml
memory: "{{ example_app_memory }}"
```

This creates a configurable default memory for containers through the docker daemon and if users want a specific app to have more mem, such as plex, or nextcloud, or jellyfin, etc. it can be configured through it respective role folder.

davestephens:
This is actually being implemented with the move to roles, which I like, so I'll close this. If someone wants to tackle any outstanding I'll happily merge!

bcurran3:
I just spot checked four roles, only one had it. At very least bitwarden, cloudmd, and duplicati still have hardcoded memory limits.

davestephens:
@bcurran3 you were right, nowhere near the coverage I had thought it was. Fixed now!

PR #254, opened 2020-04-16
Update Plex default memory limitation 1g -> 2g

bcurran3:

Changes the default memory limitation for the Plex container from 1Gb to 2Gb. (Though I prefer more, but that's addressed as a feature request in https://github.com/davestephens/ansible-nas/issues/253 .) Realigns with Plex's recommendation (see reference below). Basically more "breathing room" for transcoding.

REF: https://www.reddit.com/r/PleX/comments/61zb8m/plex_media_server_and_ram_usage/

REF: https://support.plex.tv/articles/200375666-plex-media-server-requirements/
> RAM
> In general, Plex Media Server doesn’t require large amounts of RAM. 2GB of RAM is typically more than sufficient and some installs (particularly Linux-based installs) can often happily run with even less. Of course, more RAM won’t hurt you and will certainly be helpful if you’re also doing other things on the computer.
> 

Fixes # n/a

**Any other useful info**:

### DIFF
--- a/tasks/plex.yml
+++ b/tasks/plex.yml
@@ -25,7 +25,7 @@
       PUID: "{{ plex_user_id }}"
       PGID: "{{ plex_group_id }}"
     restart_policy: unless-stopped
-    memory: 1g
+    memory: 2g
     labels:
       traefik.backend: "plex"
       traefik.frontend.rule: "Host:plex.{{ ansible_nas_domain }}"


bcurran3:
Example of my current real world use with ONE video playing:

image

davestephens:
Funnily enough I just did this on my own Plex setup...

PR #255, opened 2020-04-16
Change samba_shares_root and nfs_shares_root default to more common default

bcurran3:

DISCRETIONARY

Change samba_shares_root: to more common occurrence; md0 instead of Volume3.

Fixes # n/a

**Any other useful info**:

It's in this user's opinion that "Volume3" is a more personal reference to Ansible-NAS's author's setup and the default Samba share example might be better served using the default first mdadm RAID mount point of /mnt/md0. Of course there's the growing popularity of ZFS as well, but I don't use that. :)

This PR can be changed or shot down without prejudice.

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -147,7 +147,7 @@ docker_storage_driver: overlay2
 ###
 # The location where all shares will be created by default. Can be overridden on a per-share basis.
 # This path will be mounted to backup containers, Duplicati
-samba_shares_root: /mnt/Volume3
+samba_shares_root: /mnt/md0
 
 # Where stuff downloaded will be stored
 downloads_root: "{{ samba_shares_root }}/downloads"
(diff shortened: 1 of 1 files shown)


bcurran3:
Another "The command "mkdocs build" exited with 1. " failure...

Not sure how to re-test, or if even possible. (I looked).

bcurran3:
Missed changing the nfs_shares_root, corrected!

davestephens:
My main issue with changing this now is backwards compatibility - I'm guessing there are likely people out there that haven't bothered to change the defaults from my volume2/volume3 stuff (which is like that because volume1 died whilst creating Ansible-NAS :laughing:). 

Also, Ansible-NAS recommends using ZFS. Having defaults that are relevant to mdadm which isn't mentioned anywhere, doesn't really make much sense imo. 

I'm happy to revisit if the docs are beefed up to include mdadm as an alternative to ZFS, but for now I'm going to close this one.

bcurran3:
I completely overlooked backwards compatibility, got it.

Maybe a simple compromise is for me to through in a comment with examples such has md0 and whatever the zfs default mount is (let me know!).

PR #259, opened 2020-04-22
Adds ability to share user home dirs

bcurran3:

Adds ability to share user home dirs.

Fixes # n/a

**Any other useful info**:

Bill continues to Rage against the NAS! 
[CCC - Continous Comedic Callback](https://en.wikipedia.org/wiki/Callback_(comedy))

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -179,6 +179,15 @@ photos_root: "{{ samba_shares_root }}/photos"
 # The description that'll appear next to your Ansible-NAS box when browsing your network
 samba_server_string: Ansible NAS
 
+# Your password for samba shares that are not public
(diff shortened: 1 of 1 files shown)


bcurran3:
BTW: I struggled very much on the best place to put this. I moved it around at least three times before settling on where it is.

davestephens:
I like the idea of this, but what do you think of providing a way to add additional users to the Ansible-NAS box? We could then integrate that with the samba_users stuff.

bcurran3:
Well this shares the default ansible-nas user's home dir.  Basically makes it easier for the principle user to access his/her files remotely over a share.

Adding more users, that seems like a different PR to me.

I'm waiting for the current migration to roles to settle and then I've got some ideas... MAYBE tackle this one.

Issue #65, opened 2019-03-09
Gotify self-hosted push notifications


davestephens:
Could be used for notifications generated by Ansible-NAS, check out https://github.com/gotify/server.

bcurran3:
REF: https://hub.docker.com/r/gotify/server

I'm looking into this.

bcurran3:
I've done this one. I'll get the PR in later tonight or tomorrow.

bcurran3:
Done.

https://github.com/davestephens/ansible-nas/pull/194
https://github.com/davestephens/ansible-nas/pull/195
https://github.com/davestephens/ansible-nas/pull/196
https://github.com/davestephens/ansible-nas/pull/197
https://github.com/davestephens/ansible-nas/pull/198
https://github.com/davestephens/ansible-nas/pull/199

bcurran3:
New ORGANIZED PR done:
https://github.com/davestephens/ansible-nas/pull/264

PR #264, opened 2020-04-23
ADDITION: Gotify

bcurran3:

Add Gotify app.

Fixes # https://github.com/davestephens/ansible-nas/issues/65

**Any other useful info**:

Nice for server specific messages to SysOp.


bcurran3:
I don't think this is a "me thing."

> ERROR! Unable to retrieve file contents
> Could not find or access '/etc/ansible/playbooks/playbook_under_test/tasks/gotify.yml' on the Ansible Controller.

bcurran3:
Found the problem. Duh!

The rage shall continue.

docs/applications/gotify.md
```diff
@@ -0,0 +1,20 @@
...
+
+Gotify is a simple server for sending and receiving messages.
+
+## Usage
+
+Set `gotify_enabled: true` in your `inventories/<your_inventory>/nas.yml` file.
+
+If you want to change the default port, edit gotify_port_http:.
+
+The Gotify web interface can be found at http://ansible_nas_host_or_ip:8077 and http://ansible_nas_host_or_ip:8078 (internally).
```
davestephens, line 15: What does "8078 internally" mean, isn't that the https port?

group_vars/all.yml
```diff
@@ -823,4 +826,14 @@ vd_users:
...
\ No newline at end of file
+vd_rdp_port: 3389
+
+###
+### Gotify
+###
+gotify_available_externally: "false"
+gotify_data_directory: "{{ docker_home }}/gotify/config"
+gotify_port_http: "8077"
+gotify_port_https: "8078"
```
davestephens, line 837: This port doesn't appear to be mapped to the container?

tasks/gotify.yml
```diff
@@ -0,0 +1,28 @@
+---
+- name: Gotify Directory
+  file:
+    path: "{{ item }}"
+    state: directory
+  with_items:
+    - "{{ gotify_data_directory }}/app/data"
```
davestephens, line 7: It feels a little clunky adding `/app/data` to the end of the path, as data is then stored in `{{ docker_home }}/gotify/config/app/data`

tasks/gotify.yml
```diff
@@ -0,0 +1,28 @@
...
+
+- name: Gotify Docker Container
+  docker_container:
+    name: gotify
+    image: gotify/server:latest
+    pull: true
+    volumes:
+      - "{{ gotify_data_directory }}/data:/app/data:rw"
```
davestephens, line 15: Should drop `/data` from the path per my comment above

davestephens (changes requested):
Looking forward to getting this one merged :+1: Needs a little bit of tweaking then we're good to go.

tasks/gotify.yml
bcurran3: Intentional. (Like the other reference.) I like to make the dirs match in the config directory to the container. Personal preference only and I'm not highly committed to it.

group_vars/all.yml
bcurran3: Fixed - yeah this was pushed when I didn't have Traefik setup yet. :)

bcurran3:
Good now IMHO thought you might still want a simple change.
This one's for you - me giving back.

tasks/gotify.yml
bcurran3: BTW: NO config dir.
I'm going to start getting away from the template docker_data/app/config default and go with docker_data/app as the "config"/"data" dir. I really like matching the container's absolute path to a path under docker_data/app - visually much easier to see what's going on and debug when needed.

davestephens:
Closing in favour of #429 as discussed

PR #267, opened 2020-04-23
ADDITION: grocy

bcurran3:

Adds grocy, the self-hosted groceries and household management solution.

Fixes # n/a

**Any other useful info**:

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -77,6 +77,9 @@ bitwarden_enabled: false
 # Finance
 firefly_enabled: false
 
+# grocy
+grocy_enabled: false
+
 # Wallabag
 wallabag_enabled: false
 
@@ -727,6 +730,13 @@ bitwarden_allow_signups: false
 firefly_available_externally: "false"
 firefly_data_directory: "{{ docker_home }}/firefly"
 firefly_port: "8066"
+
+###
+### grocy
+###
+grocy_available_externally: "false"
+grocy_data_directory: "{{ docker_home }}/grocy"
+grocy_port_http: "9283"
 
 ###
 ### Nzbget
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -736,6 +736,8 @@ firefly_port: "8066"
 ###
 grocy_available_externally: "false"
 grocy_data_directory: "{{ docker_home }}/grocy"
+grocy_user_id: "1000"
+grocy_group_id: "1000"
 grocy_port_http: "9283"
 
 ###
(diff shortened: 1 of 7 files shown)


group_vars/all.yml
```diff
@@ -728,6 +731,15 @@ firefly_available_externally: "false"
 firefly_data_directory: "{{ docker_home }}/firefly"
 firefly_port: "8066"
 
+###
+### grocy
+###
+grocy_available_externally: "false"
+grocy_data_directory: "{{ docker_home }}/grocy"
+grocy_user_id: "1000"
```
davestephens, line 739: Shouldn't this be `0`?

group_vars/all.yml
```diff
@@ -728,6 +731,15 @@ firefly_available_externally: "false"
...
+grocy_group_id: "1000"
```
davestephens, line 740: And this?

tasks/grocy.yml
```diff
@@ -0,0 +1,29 @@
...
+    name: grocy
+    image: linuxserver/grocy:latest
+    pull: true
+    volumes:
+      - "{{ grocy_data_directory }}/config:/config:rw"
+    ports:
+      - "{{ grocy_port_http }}:80"
+    env:
+      TZ: "{{ ansible_nas_timezone }}"
+      PUID: "grocy_user_id"
```
davestephens, line 21: Looks like some missing `{{ }}` here

tasks/grocy.yml
```diff
@@ -0,0 +1,29 @@
...
+      PGID: "grocy_group_id"
```
davestephens, line 22: And here

group_vars/all.yml
bcurran3: From what I remember this is one of those containers that you can't run as root due to the web server.

group_vars/all.yml
bcurran3: From what I remember this is one of those containers that you can't run as root due to the web server.

tasks/grocy.yml
bcurran3: Good catch - thanks. Will fix NOW!

tasks/grocy.yml
bcurran3: Good catch - thanks. Will fix NOW!

bcurran3:
I think GTG now.

bcurran3:
> Funnily enough I had Grocy open in a tab on my mobile - was nice to see this come in!
> 
> Needs a few tweaks then we should be good to go 👍

Funny. Great minds think alike? Or scary? LOL!

davestephens:
If this is flipped to an Ansible role and tidied up I can review and merge.

bcurran3:
Thanks for the response to my statement saying it was ready to go 834 days ago. 
I'll close now and probably re-submit in < 834 days from now.

PR #268, opened 2020-04-23
ADDITION: uTorrent

bcurran3:

Adds uTorrent, providing an alternative to Transmission.

Fixes # n/a

**Any other useful info**:

My preference over Transmission.


group_vars/all.yml
```diff
@@ -368,6 +369,17 @@ openvpn_password: secretpassword
...
 
+###
+### uTorrent
+###
+utorrent_available_externally: "false"
+utorrent_config_directory: "{{ docker_home }}/utorrent/config"
+utorrent_download_directory: "{{ downloads_root }}"
+utorrent_port_http: "8111"
+utorrent_port_bt: "6881"
+utorrent_user_id: "999"
```
davestephens, line 380: What uid is this?

group_vars/all.yml
```diff
@@ -368,6 +369,17 @@ openvpn_password: secretpassword
...
+utorrent_group_id: "1001"
```
davestephens, line 381: And gid?

bcurran3: Typo, smartypants!

group_vars/all.yml
bcurran3: Jumping on it!

group_vars/all.yml
bcurran3: I think you're looking at an old commit. I just went to edit the file and it has utorrent_user_id: "999"
utorrent_group_id: "1001"
same as mine.

group_vars/all.yml
bcurran3: Sorry for my total confusion. I see what you're talking about now. I have no idea now where I got that as it's the docker id and it's not what's referenced in the docker container setup guide: their default is 1000 for both.
I'll set it to 1000 now. If you think 0 is better, feel free to change it.

group_vars/all.yml
bcurran3: Sorry for my total confusion. I see what you're talking about now. I have no idea now where I got that as it's the docker id and it's not what's referenced in the docker container setup guide: their default is 1000 for both.
I'll set it to 1000 now. If you think 0 is better, feel free to change it.

group_vars/all.yml
bcurran3: I changed it to "0" to match the Transmission defaults to keep the two aligned

group_vars/all.yml
bcurran3: I changed it to "0" to match the Transmission defaults to keep the two aligned

group_vars/all.yml
davestephens: The UID you specify needs to match one on the host system, so what they say/recommend is totally irellevant. 0 is good though 👍

PR #269, opened 2020-04-23
ADDITION: YouTubeDL-Material

bcurran3:

Adds YouTubeDL-Material, a great and beautiful YouTube downloader.

Fixes # n/a

**Any other useful info**:

I LOVE this! You will too.

Rage against the NAS!


davestephens:
:+1: Thanks for looking into the bug too!

Issue #144, opened 2019-10-15
Add SPRUT.io (web based file manager) docker container support


bcurran3:
I haven't tried it yet, but this looks pretty "nifty" and might be a good addition. After all, who doesn't need some sort of file management for their NAS?

https://sprut.io/en/

bcurran3:
I'm closing this as under further investigation the sprutio docker only has one release version and it was from 2017.

I'm looking for a Cloud Commander alternative that is more modern and will open an new issue (most likely a PR) when I find something I like.

PR #270, opened 2020-04-23
ADDITION: Krusader

bcurran3:

Krusader adds an alternative to using Cloud Commander to ansible-nas. Additionally it provides access via VNC.

Replaces my naive request for SPRUT.io: https://github.com/davestephens/ansible-nas/issues/144

Fixes # n/a

**Any other useful info**:
I don't use Guacamole. I have no idea how the VNC portion of this will interact or conflict with Guacamole.

Rage against the NAS to be continued...


davestephens:
Looks good!

I made a few changes: 
 - Disabled HTTPS to prevent certificate errors
 - Fixed the docs URL and VNC port
 - Renamed `krusader_data_directory` to `krusader_config_directory`
 - Updated the config path mounted on the container

bcurran3:
Thanks for the assist.
Please double check - "{{ krusader_config_directory }}/krusader" in the task, looks to me it will make dockerdata/krusader/config/krusader

PR #303, opened 2020-04-26
update Traefik image to latest 1.7.x release

bcurran3:

update Traefik image to latest 1.7.x release

Fixes # n/a

**Any other useful info**:

I've been running this since 04/17/20 without problems.


davestephens:
Using tag `1.7` should reference the latest 1.7 cut (which is 1.7.24 at the moment), according to this: https://hub.docker.com/_/traefik.

By changing to something more specific we're pinning to that minor version, which is not what we want.

bcurran3:
GTK: lesson learned!

PR #313, opened 2020-05-04
ADDITION: Monitorr

bcurran3:

Adds Monitorr

Fixes #

**Any other useful info**:

goes well with Organizr


davestephens:
Looks like the developer has given up on this. If you can find something otherwise then let me know...

bcurran3:
It was fresh and shiny when I submitted it 845 days ago. Thanks for responding, better 2 years late than never.

I have this installed but never use it. I don't see any need to resubmit.

PR #315, opened 2020-05-05
BUG: Transmission UDP fix

bcurran3:

Exposes the UDP port, same as the TCP port (transmission_external_port/transmission_openvpn_external_port), to allow DHT use.

Fixes # n/a

**Any other useful info**:
I didn't see the need to update the ports list explicitly for the UDP addition as the TCP port is already there.


davestephens:
Funnily enough I was playing with this the other week and hadn't realised it was due to the missing UDP, good spot :+1:

bcurran3:
image

PR #320, opened 2020-05-09
ADDITION: TVHeadend

YurNaybor:

This PR adds TVHeadend as an installable service via Ansible and Docker. TVHeadend is a streaming and recording server for television broadcasting and can be used with different backends, like IPTV, DVB tuner cards and so on.

 #325 

**Any other useful info**:

Though I think the PR is ready as it is (and tested locally), there are still some thoughts I have:

- I added a new directory / share for recordings, because I think the "tv" folder does not quite fit the needs of TVHeadend which not only records series. I'm open for other opinions here.
- DVB drivers / firmware: I wasn't sure whether to include tasks for installing DVB drivers and firmware. For my personal hardware (WinTV Quad HD), I would need, for example, the Hauppauge ppa and the Linux media tree (see https://hauppauge.com/pages/support/support_linux.html#ppa). As this is very specific to the users hardware, I currently see this out of scope for ansible-nas.

### DIFF
--- a/docs/applications/tvheadend.md
+++ b/docs/applications/tvheadend.md
@@ -22,3 +22,9 @@ will then add the device / folder '/dev/dri' to the list of docker volumes. This
 
 If you want to pass a physical dvb tuner card, you can set `tvheadend_hw_dvb` to `true`. The play 
 will then add the device / folder '/dev/dvb' to the list of docker volumes. This is disabled by default.
+
+For discovery of local IPTV services, you may need to change the network mode of the docker container to
+ `host`, default is 'bridge'.
+
+It is possible to pin the version of the Docker images by setting `tvheadend_version` to the desired tag.
+ Note that the linuxserver.io images are used for TVHeadend.
--- a/docs/applications/tvheadend.md
+++ b/docs/applications/tvheadend.md
@@ -1,0 +1,24 @@
+# TVHeadend
+
+Homepage: [https://tvheadend.org/](https://tvheadend.org/)
+
+Tvheadend is a TV streaming server and recorder for Linux, FreeBSD and Android supporting DVB-S, 
+DVB-S2, DVB-C, DVB-T, ATSC, ISDB-T, IPTV, SAT>IP and HDHomeRun as input sources.
+
+## Usage
+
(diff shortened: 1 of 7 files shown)


docs/applications/tvheadend.md
```diff
@@ -0,0 +1,30 @@
...
+setting `tvheadend_recordings_directory`.
+
+If you want to enable hardware acceleration, you can set `tvheadend_hw_accel` to `true`. The play 
+will then add the device / folder '/dev/dri' to the list of docker volumes. This is disabled by default.
+
+If you want to pass a physical dvb tuner card, you can set `tvheadend_hw_dvb` to `true`. The play 
+will then add the device / folder '/dev/dvb' to the list of docker volumes. This is disabled by default.
+
+For discovery of local IPTV services, you may need to change the network mode of the docker container to
+ `host`, default is 'bridge'.
```
Vinibo, line 27: :princess: For clarification, I would add the full property name `tvheadend_network_mode`.

tasks/tvheadend.yml
```diff
@@ -0,0 +1,51 @@
...
+- name: Define basic docker devices
+  set_fact:
+    tvheadend_docker_devices: []
+
+- name: Add hw acceleration
+  set_fact:
+    tvheadend_docker_devices: "{{ tvheadend_docker_devices + [ '/dev/dri:/dev/dri' ] }}"
+  when: tvheadend_hw_accel
+
+- name: Add dvb hw
```
Vinibo, line 11: This is a little 'princess' comment but it seems a bit cryptic with two abbreviations in the name. Maybe something along the lines of 'Add hardware TV Tuner' or 'Mount hardware TV Tuner' would be a little more accessible.

Vinibo:
I went on and tested it on my 'test' server. I don't know what caused it but when I complete the 'Quick setup', I end up locked out of the TVHeadend interface with a password prompt. I tried the 'admin' and 'user' password specified during the quick setup but nothing seems to work.

Here's my configuration:
```
tvheadend_enabled: true
tvheadend_network_mode: host
```
I had to put 'tvheadend_network_mode: host' for the discovery of my HDHomeRun.

davestephens:
Closing due to no activity, and I don't have the hardware to test it.

PR #324, opened 2020-05-14
Custom subdomain

### LINKED ISSUE #109: Allow configurable subdomains

StopMotionCuber:

Basically this is a duplicate/new revision of #201 as that PR was stale for a long time.

This PR adds support to customize the suburl of services that are made available externally.
It also adds support to only request certificates for services that are actually active.
This makes ACME (without DNS) possible even if you don't have all the subdomains registered.

Fixes #109

**Any other useful info**:

What do you think of this PR?
Do we need any additional stuff besides documentation for this?

I worked much with regex replacement, so please tell me if I have overseen something that is not formatted according to the regex I used

Note: The [sans] value of traefic.toml will not be formatted in a nice way, do we need that?
Working with jinja and trying to get good whitespacing there is not quite fun, and as the output is correct, this should be sufficient (in my opinion)


StopMotionCuber:
I added some documentation now.

Note that my PR has no documentation yet on the functionality that only the enabled SANs are requested. Also there seems to be an issue with re-requesting new certificates on configuration changes (new subdomain, new SANs) that either should get
 - Documented
 - Workaround
 - Potentially both
I'm thinking here of placing the `acme.json` in a folder that mounted into the container.
Then the user has the possibility to remove that certificate to force the creation of a new certificate. But, as I said, I still need to do some testing on that (maybe there is an option that requires less user interaction).

StopMotionCuber:
I think the current state of customizing subdomains is mergeable.

Note that I removed the section about `Letsencrypt will only allow a total of 5 duplicate certificates to be issued in a rolling 7 day period`.
While this is correct, this information is misleading: "A certificate is considered a renewal (or a duplicate) of an earlier certificate if it contains the exact same set of hostnames, ignoring capitalization and ordering of hostnames. For instance, if you requested a certificate for the names [www.example.com, example.com], you could request four more certificates for [www.example.com, example.com] during the week. If you changed the set of hostnames by adding [blog.example.com], you would be able to request additional certificates." [https://letsencrypt.org/docs/rate-limits/](https://letsencrypt.org/docs/rate-limits/)

This means, changing (or adding) services one-by-one would still lead to hit that limit, but as these requests contain different sets of domains, they are not considered duplicates.

@davestephens What do you think about this?
@bcurran3 I think I already need to apologize in advance, if this will get merged, it will break half of your PRs, as they are based on hardcoded subdomains

bcurran3:
@StopMotionCuber - HA! No worries. I'll update whatever needs to be updated. I've got another PR that will require modifying all but a few tasks and am holding off until @davestephens catches up on his apPRoving.

davestephens:
Sorry to do this - with the flip to Traefik 2.4 and move to roles, this PR has become obsolete to a point where it'll be a nightmare to merge. If it's ok with you I'll close this and do this work as I move the remaining tasks into roles 👍

PR #328, opened 2020-05-17
CONTAINER CHANGE: Watchtower to use creator/maintainer's container

bcurran3:
**What this PR does / why we need it**:

Current container used (v2tec/watchtower) hasn't been updated in two years. 
Meanwhile, containnrrr/watchtower, is the developer's container and updated regularly (as recent as a week ago). Also supports more notifications (e.g. Gotify and another just got merged into the watchtower codebase recently as well).

**Which issue (if any) this PR fixes**:

Fixes # n/a

**Any other useful info**:

GOTIFY SUPPORT!

PR #329, opened 2020-05-17
Containrrr/watchtower   simpler and granular configuration

bcurran3:

Easier to granularly configure via ENV instead of long command string.

Fixes # n/a

**Any other useful info**:

REQUIRES: container change from https://github.com/davestephens/ansible-nas/pull/328

DISCLAIMER: I have NOT tried this with the notification setting to "" (blank) - need to test to make sure it doesn't generate an error; e.g. defining POLL_INTERVAL as "" and using scheduling will generate a "Only schedule or interval can be defined, not both." error from Watchtower.


bcurran3:
A visual to maybe add enthusiasm:

image

davestephens:
Thank you but I'm going to pass on this one - I actually prefer the command based config for this container due to the insane amount of options it supports. Furthermore, by keeping it as a command rather than parameterised options, there are no requirements to update the task defs if more configuration options are added upstream.

PR #331, opened 2020-05-19
fix video path

xerosanyam:
I felt the video path had a typo

Fixes #

**Any other useful info**:

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -553,7 +553,7 @@ lidarr_port: "8686"
 youtubedlmaterial_available_externally: "false"
 youtubedlmaterial_data_directory: "{{ docker_home }}/youtubedlmaterial"
 youtubedlmaterial_dl_audio_directory: "{{ downloads_root }}/youtube/audio"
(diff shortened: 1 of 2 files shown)


bcurran3:
@xerosanyam - you are 100% right. youtubedlmaterial_dl_video_directory: "{{ downloads_root }}youtube/video" is missing a slash. Since you closed this PR, I just put in one to fix it: https://github.com/davestephens/ansible-nas/pull/332.

As to where you want the videos to download to (non-default), I'm guessing you already realized that you can change it to any place you want. Again, thanks for pointing out the error in the default!

p.s. When I added this I put a lot of thought into where the videos SHOULD download to and I don't think there's a one size fits all solution. For me, since the files are downloaded, I wanted the videos to go to my downloads directory to then curate (view, sort, move, delete, etc). This very well could be a "me thing." 

You know you can change the default easily for you personally, but if you have a strong feeling the default for Joe User should be in videos_root, open an issue and request a vote; I won't take it personally. :)

xerosanyam:
hey @bcurran3, I closed the pr because added some unintentional extra code. 

you are right, the paths are user-specific. I understand it now :)
thanks for taking this.

PR #327, opened 2020-05-17
DOCS: Fixed link for YouTubeDL-Material

mjmurphy:
**What this PR does / why we need it**:
Update to Readme.md
- the YouTubeDL-Material was linking to the [youtube-dl](https://ytdl-org.github.io/youtube-dl) project.
- modified text on Krusader link

PR #352, opened 2020-07-23
Fix YoutubeDL-Material links

Tzahi12345:
Updated link to YoutubeDL-Material to the actual link (it was set to the youtube-dl project before)

It fixes the link for YoutubeDL-Material, it was previously set to the youtube-dl home page. youtube-dl is the command-line utility that YoutubeDL-Material is based on, so the confusion is very understandable.

N/A

**Any other useful info**:

N/A


bcurran3:
Please see also https://github.com/davestephens/ansible-nas/pull/327
Yours does address both link locations though. 👍

Tzahi12345:
Ah ok, thanks for the note! I'll leave it up to the author whether to merge or not, I don't mind either way :)

davestephens:
Thanks! 👍

PR #336, opened 2020-05-25
replaced cloudflare-ddns

odbaeu:
**What this PR does / why we need it**:
I ran into bug #243 and tried to fix it. After some research I replaced the docker container with cloudflare-ddns from [oznu](https://hub.docker.com/r/oznu/cloudflare-ddns), which seems to be more popular (10M+ downloads vs. 5M downloads on docker hub). But for me it was much more important that the code was easier to understand.

**Which issue (if any) this PR fixes**:
Fixes #243

**Any other useful info**:
`network_mode: host` is required for IPv6 to work properly.

PR #354, opened 2020-08-13
Cloudflare-ddns: fix broken configuration and update default auth from deprecated Global API Key to API Token 

### LINKED ISSUE #243: Upstream Bug: cloudflare-ddns no longer working
The Cloudflare DDNS image joshava/cloudflare-ddns has not been updated in over a year - I'm unsure if it is abandonware or not. Cloudflare have depreciated global API calls, instead you now issue an API key for individual domains and actions. You can no longer issue the global API key the docker image needs. This leaves the docker image no longer working for new installs and if it still works for existing installs, it won't do for much longer.

As a sidenote, we are also not using cloudflare effectivelyas we simply issue a wildcard subdomain pointing to our install. On free plans this means we cannot proxy the subdomains without either manual setup or paying cloudflare silly monry to upgrade the plan.

ryanolf:

This PR fixes #243 for current configurations. Somehow, for me, even the latest joshava/cloudflare-ddns image with support for the latest Cloudflare API was not working for me. The author of that docker image has deprecated configuration by environment variables, so I refactored the ansible task to use a config file.

As Cloudflare has deprecated global API key access, I also updated the ansible-nas default to use the API Token authentication. To use API Token auth, existing users need to set the appropriate variable and remove the `cloudflare_api_key` variable. However, existing users need not change their configuration at all, so long as Cloudflare supports the global API key. 

Fixes #243 

**Any other useful info**:

I have tested this both with my old configuration (using global API key) and with a new access token. I believe this PR should be a straightforward fix for #243 (along with, obviously, the fact that the docker image was updated) and should allow ansible-nas users to move on from the deprecated global API key to the more granular and secure API Tokens.

### DIFF
--- a/docs/applications/cloudflare_ddns.md
+++ b/docs/applications/cloudflare_ddns.md
@@ -12,4 +12,6 @@ Set `cloudflare_ddns_enabled: true` in your `inventories/<your_inventory>/nas.ym
 
 ## Specific Configuration
 
-Make sure you set your Cloudflare login, domain and API key details within your `inventories/<your_inventory>/nas.yml` file.
+Make sure you set your domain (if different than the ansible-nas default) and access token details within your `inventories/<your_inventory>/nas.yml` file. If you need to create an API token, see https://joshuaavalon.github.io/docker-cloudflare/guide/cloudflare.html#authentication for instructions.
+
+Cloudflare has deprecated global API key authentication. If you have an older ansible-nas configuration using a global API key, you can upgrade to the API token-based authentication by removing the `cloudflare_api_key` variable from your local `nas.yml` configuration file and setting the `cloudflare_token` variable appropriately.
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -300,6 +300,8 @@ cloudflare_host: "*.{{ cloudflare_zone }}"
 cloudflare_token: ""
 
 # Set to true to make traffic go through the CloudFlare CDN.
+# Note that if the cloudflare host is a wildcard (the default), this must be false, as cloudflare
+# does not support http proxy of wildcard CNAMEs.
 cloudflare_proxy: false
 
 # Set to AAAA to use set IPv6 records instead of IPv4 records.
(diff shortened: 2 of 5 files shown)


ryanolf:
FYI to all, this pull request conflicts with #336 . That PR, I believe, would break/not fix existing configurations.

@davestephens I don't know if Cloudflare-DDNS is still broken for everyone (it was for me, even with the updated joshuaavalon docker image), but if it is, merging this PR should provide immediate relief without any config updates. And those using/wanting to use Token Auth with have a path forward as well with some minor effort.

davestephens:
Hey, thanks for the ping.

I know I'm horrifically behind on PRs, apologies for that. A new job and
kids at home has massively reduced my time for hacking recently! I hope to
embark on a clear out over the next couple of weeks. Additionally, I have a
rejig planned that'll make merging PRs much much much easier 👍

odbaeu:
I can confirm that this PR is working with the new authentication method (api token) of cloudflare.
Here is my configuration:
```yml
###
### Cloudflare
###
# Cloudflare is a great free DNS option for domains. If you use the cloudflare_ddns container then you'll need to
# set the options below.
# Data directory for config file
cloudflare_data_directory: "{{ docker_home }}/cloudflare_ddns"

# Your domain name
cloudflare_zone: "{{ ansible_nas_domain }}"

# The hostname you want the container to update. You shouldn't need to change this.
cloudflare_host: "{{ ansible_nas_hostname }}.{{ cloudflare_zone }}"

# Cloudflare scoped token (https://joshuaavalon.github.io/docker-cloudflare/guide/cloudflare.html#authentication)
# Make sure token permissions include #DNS:Edit and #Zone:Read
cloudflare_token: "***"

# Set to true to make traffic go through the CloudFlare CDN.
# Note that if the cloudflare host is a wildcard (the default), this must be false, as cloudflare
# does not support http proxy of wildcard CNAMEs.
cloudflare_proxy: false

# Set to AAAA to use set IPv6 records instead of IPv4 records.
cloudflare_type: "A"
```

davestephens:
I finally got around to merging this, thank you for the contribution!

Couple of notes:
- I flipped your code to using the new ansible role structure
- I took the executive decision to not support the global cloudflare api keys, thus making the code much cleaner :-)
- I standardised on `.yml` over `.yaml`.

PR #355, opened 2020-09-07
Update portainer.yml

rodrigorodrigo:

updates Portainer package from portainer/portainer to portainer/portainer-ce, updating it to version 2.0.

nothing really (other than updating portainer container)

**Any other useful info**:


davestephens:
Thanks for the PR! 👍

PR #357, opened 2020-09-30
Nextcloud MySQL credentials changed to variable for the security paranoid.

Blackfell:

Very small pull request to change the MySQL creds from a static string to variable that can be overridden by the user if desired. Whilst a minor security change, something I've been doing when exposing my Nextcloud instances to the internet as a potential damage limitation piece.

Default behavior remains as before, only affects implementation if overridden by the user. 

None.

Fixes #

**Any other useful info**:

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -107,7 +107,7 @@ pytivo_enabled: false
 serposcope_enabled: false
 
 # External Access
-# Traefik will allow access to certain applications externally. To enable this you'll either; a domain name that points to your 
+# Traefik will allow access to certain applications externally. To enable this you'll either; a domain name that points to your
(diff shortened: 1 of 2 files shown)


davestephens:
Thanks for the contribution! I like the idea, but you've (accidentally?)  changed the defaults which will break anyone with an existing install. If you can set the default values to what they were, I'll get this merged :+1:

Blackfell:
Thanks and well spotted! You're right, copy paste error *facepalm* 

That should all be back to default now, thanks for helping me out with it :)

davestephens:
Finally got this merged, thanks for the contribution!

Blackfell:
More than happy to help, thanks for the project!

PR #362, opened 2020-10-15
Plex: Enable some advanced features of docker container, including hardware transcode

ryanolf:

There are features of the Plex docker container that at present are not accessible from ansible-nas. Notably, I wanted to enable hardware transcode, which requires mapping a device into the docker container (as described in the [container docs](https://hub.docker.com/r/linuxserver/plex)).

The change here is minor: it allows a device mapping variable to be set, which can be used to enable hardware transcoding and dab support. In addition, a "plex_version" is used to allow easier upgrades, another feature of the container. The default setting in all.yml keeps the current behavior.

With this PR, others can easily enable hardware transcoding, dab support, or beta-channel updates by updating their config variables. If folks do not modify their configs, this PR does not change anything.

**Any other useful info**:

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -436,6 +436,10 @@ plex_music_permissions: "rw"
 plex_user_id: "0"
 plex_group_id: "0"
 plex_port: "32400"
+plex_version: "docker"
+# Device mappings for the docker container.
+# E.g. To enable hardware transcoding:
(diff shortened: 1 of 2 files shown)


davestephens:
Hey @ryanolf - I incorporated your changes in this commit https://github.com/davestephens/ansible-nas/commit/8d5cfc9dc0dfb482eedf0860d406c5d5a380bbe4 - thanks for the help! 👍

PR #364, opened 2020-10-28
More consistent use of variables in Samba share paths

elliotsayes:
Updates the template all.yml file to use existing path variables instead of redefining the paths.


davestephens:
Good spot!

PR #372, opened 2020-12-30
Addition - Nodemation, n8n

Narcolapser:

Effectively a selfhosted IFTTT, just with a lot more features.

None

**Any other useful info**:

Have you considered setting up a system for unifying admin user name and password across applications? I see a couple pull requests for MySQL password as a variable, taking that idea to the extreme and applying it to everything. I mention this because this application has a default user name and password, and when you throw it up for external access that is a potential security flaw.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ Ansible config and a bunch of Docker containers.
 * [Mosquitto](https://mosquitto.org) - An open source MQTT broker
 * [Mylar](https://github.com/evilhero/mylar) - An automated Comic Book downloader (cbr/cbz) for use with SABnzbd, NZBGet and torrents
 * [MyMediaForAlexa](https://www.mymediaalexa.com/) - Lets you stream your music collection to your alexa device
+* [n8n](http://n8n.io/) - Nodemation, a node based workflow and automation service.
 * [Netdata](https://my-netdata.io/) - An extremely comprehensive system monitoring solution
 * [Nextcloud](https://nextcloud.com/) - A self-hosted Dropbox alternative
 * [NZBget](https://nzbget.net/) - The most efficient usenet downloader
--- a/docs/applications/n8n.md
+++ b/docs/applications/n8n.md
@@ -1,0 +1,17 @@
+# Nodemation (n8n)
+
+Homepage: [ihttps://n8n.io](https://n8n.io)
+
+
+## Usage
+
+Set `n8n_enabled: true` in your `inventories/<your_inventory>/nas.yml` file.
+
+Set `n8n_secured: "true"` if you want to limit access to the workflows. The user and password can be set with:
+* n8n_basic_auth_user: "<user name>"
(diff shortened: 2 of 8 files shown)


davestephens:
Hey, thanks for your contribution!

This is a great start. 

Few comments:
- Please can you remove the changes to the nextcloud.yml
- Why is there two definitions of the docker container, then a conditional available_externally and whether n8n is secured or not? There definitely doesn't need to be on the available_externally variable, and from a cursory glance, the n8n_basic_auth variable doesn't require it either?

Narcolapser:
The nextcloud changes, I've been trying to, didn't mean to include them to begin with, but you'll notice in commit fef38d9d I did remove the trailing white space but it still shows it as being there. Notice in commit 1686532 that I appear to have made no changes. So far as I can tell nothing is actually changed in that file, but if it matters we can cancel this PR, and I'll remake my changes this time without touching nextcloud. 

As for the two containers, options: I'm still new to ansible so correct me if I am wrong here, but my understanding is that the task has the conditional. If I could put the conditional on parts of the task, like the environment variables, then that would be more elegant. But I don't think that's what you are saying, it sounds like you we don't need to put in the conditional. The reason I put in the conditional is to force auth if you expose your node instance, since it is completely open by default. This is a huge security hazard since it allows the running of arbitrary cli commands. Really I should find a way to make a completely random password by default considering how much of a security hazard opening your machine to arbitrary commands is.

davestephens:
- Please remove the changes to `nextcloud.yml`.
- Please remove lines 9-21 in `n8n.yml.
- Create a dictionary in all.yml:
```
    n8n_environment_variables:
      N8N_BASIC_AUTH_ACTIVE: "true"
      N8N_BASIC_AUTH_USER: "{{ n8n_basic_auth_user }}"
      N8N_BASIC_AUTH_PASSWORD: "{{ n8n_basic_auth_password }}"
```
change `env:` in the `docker_container` definition to:
```
env: "{{ n8n_environment_variables | default(omit) }}"
```

Then, if `n8n_environment_variables` is not defined, the env: param isn't passed, thus saving on cutting and pasting the same docker_container code into the task yml. You'll need to add some docs to tell people to comment it out/enable it based on if they want auth. Alternatively ditch the logic and simply leave auth turned on - up to you. Many apps have auth on by default.

Narcolapser:
Can do. I'll try to commit a new PR this evening that doesn't have the next cloud edits I can't seem to get rid of.

Narcolapser:
Wait, one more question. Why a role instead? Are you moving away from tasks to roles or is there something else motivating this request?

davestephens:
By having the application in a role it makes merges easier and
configuration for the application cleaner.

I'm currently moving everything into roles.

Narcolapser:
That was what I was guessing. I agree. Do you have one I can use as an example? That way I can try to keep it in the same style.

davestephens:
For sure - check out transmission-with-openvpn.

PR #378, opened 2021-01-17
Move bitwarden, nextcloud, duplicati and homeassistant to roles

joao-p-marques:

Seems like some **tasks** are being ported into **roles** (https://github.com/davestephens/ansible-nas/commit/4cf0be93a181a5e0b4adee451635c8a2381792d1 and similar)

This moves **bitwarden, nextcloud, duplicati and homeassistant** to independent roles.


davestephens:
This is great, thank you, I appreciate the help!

Please can you move the `*_enabled` variables from `all.yml` into the
defaults file within the roles too?

joao-p-marques:
Thanks @davestephens 

Changes made. I also removed the `plex_enabled` var which I think was left over from the port to role.

Taking a closer look at the file structure, I also noticed that the `ansible_nas_user` variable was defined both in the `all.yml` file and in the `roles/ansible-nas-docker/defaults/main.yml` file. I guess it could be removed from 1 of them, right?

Also, taking a look at [the installation docs](https://davidstephens.uk/ansible-nas/installation/), in point 5 you mention:
> Review `group_vars/all.yml`. Change settings by overriding them in `inventories/my-ansible-nas/group_vars/nas.yml`. 

I guess, if the `_enable` and configuration variables end up being moved to the specific roles, the proper place to have all of them together as an example would probably be in `inventories/sample/group_vars/`, right?

davestephens:
>>  Review group_vars/all.yml. Change settings by overriding them in inventories/my-ansible-nas/group_vars/nas.yml.
>
> I guess, if the _enable and configuration variables end up being moved to the specific roles, the proper place to have all of them together as an example would probably be in inventories/sample/group_vars/, right?

The plan is that all.yml will only have "global" variables - email address, ansible-nas hostnames etc etc, all of the application specific stuff will be tucked away into roles and overrides in nas.yml.  So yeah, the installation docs will need updating once the flip to roles is complete.

That aside, thanks for the PR :)

PR #379, opened 2021-01-17
Add pre commit

joao-p-marques:
:warning: This PR involves a big diff and a potential change in workflow. I don't know if it is something that is of interest for the project maintainers, but I think it is a great improvement as it can unify code formatting easily.

Feel free to turn it down if it goes completely out of scope for what you want for the project :sweat_smile: 

This adds **pre-commit** support to the repo. 
In essence, it:
1. Adds the pre-commit (and derivates) config
2. Applies a 1st run that automatically reformats current code
3. Fixes some linting errors currently present that needed manual action:
3.1. Rename roles, replacing `-` with `_`.
3.2. Set file and directory permissions when creating them on tasks (for files, `mode: 0644` and, for folders, `mode: 0755`)
3.3. Rename the `traefik.toml` file to `traefik.toml.jinja`, to better specify that it is a template and avoid linting errors.
4. It also adds a Github action for running pre-commit on each commit, which would also allow to remove that step from the Travis config. I don't know what would be the opinion on completely migrating to GH Actions, but it also brings some advantages (such as better integration in PRs and code). 

**Any other useful info**:

The pre-commit configuration is based on https://github.com/copier-org/autopretty, which is a [copier](https://github.com/copier-org/copier) template that helps generate and maintain these configurations. If you don't know copier, it is a project scaffolding tool, which is great to keep this configs up to date and maintainable. It basically bootstraps the pre-commit config using the template (which could be left as is), but it also allows for easy updates when the template upstream changes, in which case, you only need to run `copier -a .copier-answers.autopretty.yml update` (check https://github.com/copier-org/autopretty#get-updates for more info).


davestephens:
I've not seen pre-commit before - it looks cool! 👍 

Whilst I really like the look of this, it raises a few points:
- Lots of deps are added for anyone that wants to do dev on ansible-nas. We'd need setup scripts for linux (and windows?) to get the right bits installed. Looks like node, python3, pip?
- I'd like to see the settings in this PR, and once we're agreed on what tehy should be, the actual application of them in a separate PR. 
- I do like the thought of GH actions, yeah. I've not spent any time looking at it yet though.

Thanks!

joao-p-marques:
Thank you! I am happy to bring something new to the project, then. 🙂

Pre-commit basically manages several git hooks that will be executed each time you commit to ensure the code follows some expected format and rules and, in some cases, automatically reformat. 
For example, it can run `ansible lint` before each commit, instead of only running that in the CI. 

About your questions:

> - Lots of deps are added for anyone that wants to do dev on ansible-nas. We'd need setup scripts for linux (and windows?) to get the right bits installed. Looks like node, python3, pip?

Actually, that's one of the advantages of pre-commit. It manages all the hook's installation and dependencies through a virtual environment and the developer doesn't need to worry about it. The only thing needed is:
```bash
pip install pre-commit # install pre-commit itself through pip (or pipx, better 😊)
# In the protect folder
pre-commit install # install pre-commit in the git hooks so that in can run at each commit
```
Then, pre-commit automatically installs everything in the next commit and runs each hook through the changed files (or you can run `pre-commit run -a` to run all hooks against all files). 

> - I'd like to see the settings in this PR, and once we're agreed on what tehy should be, the actual application of them in a separate PR. 

I see. However, once you install pre-commit, it will try to run in each commit next and reformat current code. So, IMHB, it makes more sense to make those changes in the same PR. I have separated them, however, in different commits: the 1st adds the configuration, the 2nd applies the auto formatting, and the 3rd applies changes that required manual action.  

> - I do like the thought of GH actions, yeah. I've not spent any time looking at it yet though.

Ok. It is a good alternative to Travis as it is built into GitHub (so, better integration) and doesn't have as many usage limitations, I think.
It is also easy to set up with a lot of "actions" you can use to easily run some steps. 

Thanks again for the feedback. I hope to have helped 😊

davestephens:
Just wanted to say thanks for the idea with this one - whilst the pre-commit config in this PR is a bit more aggressive than I want to go, pre-commit works great for the things I *do* want to do. I've started the flip over in the repo now, and will turn it on properly soon. Thanks again! 🙏

PR #382, opened 2021-01-23
N8n role

Narcolapser:
What this PR does / why we need it:

Effectively a selfhosted IFTTT, just with a lot more features.

Which issue (if any) this PR fixes:

None

Any other useful info:

I think I got it compliant with everything. You made some changes to traefik to I can see. I like the new way of doing it. Keeping as much as possible centralized in the role makes lot of sense as the number of roles continues to grow.


davestephens:
Looks good!

Please can you add the app to the list in the readme?

Also, looks like we're missing `n8n_hostname` from `defaults.yml`.

Narcolapser:
Woops. Should be good now.

davestephens:
Awesome, thanks for the contribution! 👍

PR #380, opened 2021-01-23
add deluge

PurpleNinja225:
**What this PR does / why we need it**:
Adds deluge bit torrent web client. It's my favorite torrent client and variety is nice. 

**Which issue (if any) this PR fixes**:
formatted in the new roles/{{ application }} format and fixes some of the things mentioned in a previous pull request #56 

Fixes #
#56 

**Any other useful info**:
All usefull info for operating should be in the added doc file.

PR #383, opened 2021-01-28
add deluge (re-upload of #380)

PurpleNinja225:

Adds Deluge bit torrent web client. 

Fixes #380 issues I made by not creating a separate branch on my fork.

**Any other useful info**:
I set deluge's downloads volume to be mounted at /root/downloads instead of /downloads because the default config of the client have the downloads folder pointed to that directory.


davestephens:
Awesome! 👍  Thanks for the contribution.

PR #388, opened 2021-01-29
move apps to role format

PurpleNinja225:

Rolling move of apps to the role based format. will make 1 commit/app from here going fwd.

Fixes #

**Any other useful info**:

added app template role to show example and to make moving apps to roles a bit faster (for myself at least)

##Currently moved apps in this PR:
```yml
nzbget:
pyload:
tautulli:
get_iplayer:
jackett:
```


davestephens:
Thanks! I'll try and get to this over the weekend. 👍

Going forward, it helps me with merging and review if you can keep PRs to
once piece of functionality at a time (this is actually in the contribution
docs 🙂)

PurpleNinja225:
Will do. Just added jackett cause I already had it done before seeing this :sweat_smile:

nas.yml
```diff
@@ -98,6 +98,11 @@
         - homeassistant
```
PurpleNinja225, line 98: Missing get_iplayer role

roles/application-template/defaults/main.yml
```diff
@@ -0,0 +1,21 @@
+---
```
PurpleNinja225, line 1: mv to its own PR

roles/application-template/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
+---
```
PurpleNinja225, line 1: mv to its own PR

PR #399, opened 2021-02-02
update-application_ports.md revamp of #262

PurpleNinja225:

Add a bridge/host  column to show if the application utilizes the host or bridge network by default.  Fleshes out the applications and ports a bit more.

Fixes #262 

**Any other useful info**:


davestephens:
Nice! 👍

PR #403, opened 2021-02-03
mv minidlna to roles

PurpleNinja225:

moves minnidlna to roles

Fixes #

**Any other useful info**:


davestephens:
👍

PR #406, opened 2021-02-05
fix indentation causing build failures

PurpleNinja225:

Fixes build issues in PRs #388 #404 #409 .etc

**Any other useful info**:


PurpleNinja225:
did some digging on why build failures started happening and found that I messed up the indentation for the minidlna role.

davestephens:
I actually spotted this and fixed elsewhere :-)

PR #413, opened 2021-02-06
mv miniflux to roles

PurpleNinja225:

Fixes #

**Any other useful info**:

### DIFF
--- a/roles/miniflux/tasks/main.yml
+++ b/roles/miniflux/tasks/main.yml
@@ -26,10 +26,6 @@
     pull: true
     links:
         - miniflux-postgres:db
-    volumes:
-      - "{{ miniflux_config_directory }}:/config:rw"
-      - "{{ miniflux_logs }}:/logs:rw"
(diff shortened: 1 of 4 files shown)


roles/miniflux/tasks/main.yml
```diff
@@ -18,13 +19,17 @@
...
-- name: Create Miniflux container
+- name: Create Miniflux Docker Container
   docker_container:
     name: miniflux
     image: miniflux/miniflux:latest
     pull: true
     links:
         - miniflux-postgres:db
+    volumes:
+      - "{{ miniflux_config_directory }}:/config:rw"
```
davestephens, line 30: This isn't defined in defaults.yml

roles/miniflux/tasks/main.yml
```diff
@@ -18,13 +19,17 @@
...
+      - "{{ miniflux_logs }}:/logs:rw"
```
davestephens, line 31: This isn't defined in defaults.yml too

davestephens:
👍

PR #415, opened 2021-02-06
mv calibre to roles

PurpleNinja225:

Fixes #

**Any other useful info**:


bcurran3:
Maybe we can use the move to a role as an excuse to get this renamed to Calibre-Web so that Calibre proper (linuxserver/calibre) can be added soon. Technically a breaking change but some migration could be done to handle it..

PurpleNinja225:
Makes sense since it is the calibre web image. I have a nice little template that works pretty well for most linuxserver.io docker images. Once I'm back to my laptop I'll make the change.

bcurran3:
SIDENOTE: @PurpleNinja225 - You're awesome for all your efforts!

RE: migration - my thoughts

All references/variables for "calibre" should be changed to "calibre_web" or "calibreweb." Meaning a new persistent data storage directory so that calibre proper can use a directory named "calibre." 

That's the breaking change for an existing calibre-web installs as calibre. So if there's existing data in the calibre data directory, it should be moved to the calibre_web data directory. Of course this needs to be a one-time migration to keep the program data directories seperate. There's no calibre proper container in ansible-nas yet, so it makes it difficult to compare directly at the moment, but there's got to be something unique about the old calibre-web in the calibre data directory that can be used as a trigger to move the data to a new calibre_web data directory. There's probably a unique file or directory name that is in calibre-web that would be a flag to move it from /data/calibre to /data/calibre_web that would not exist post migration. 

Sorry I can't point out that flag as i don't use calibre(-web) since I don't have calibre installed on a different computer to use it. I use Ubooquity. Once calibre can be added to ansible-nas I plan to run both.

Of course with all that said, the easy way would be just to update the docs and say running the playbook after such-and-such ansible-nas commit date will result in a new instance of calibre-web and user's should move there data from ./calibre to ./calibre_web with the container not running. But I'm guessing a little magic using Ansible's version of if exist/than work can do the trick easily for migration and not break existing installs. :)

PurpleNinja225:
People will have to migrate their container over manually if they want to keep the configs, but the nice thing about calibre-web is that it's just a pretty web front of the calibre container that is also a available and can turn on caibre-web on it's own there may be a way to install calibre and move the now named calibre_web configurations into calibre and where it stores the configs for it's calibre_web. 

I think after all the applications have been *role*d over, it would be worth looking into switching from using calibre-web to the calibre container as it has the calibre-web features plus the benefit of an accessible full calibre desktop client. Kinda jank, but could work. 

Adding new applications is pretty low on my priority, I'm mostly worried about getting apps migrated and updating the documentation, but not quite sure how Dave wants to make those changes atm I opened a discussions about where and what needs to be updated, so I can at least start making the most obvious changes.

davestephens:
Tackled a while back as part of a larger move

PR #421, opened 2021-02-07
move znc to role

PurpleNinja225:

Fixes #

**Any other useful info**:

### DIFF
--- a/roles/znc/tasks/main.yml
+++ b/roles/znc/tasks/main.yml
@@ -15,7 +15,6 @@
     volumes:
       - "{{ znc_data_directory }}:/znc-data:rw"
       - "/etc/timezone:/etc/timezone:ro"
-    network_mode: "bridge"
     ports:
       - "{{ znc_port }}:6677"
     env:
--- a/roles/znc/tasks/main.yml
+++ b/roles/znc/tasks/main.yml
@@ -3,10 +3,11 @@
   file:
     path: "{{ item }}"
     state: directory
+    # mode: 0755
   with_items:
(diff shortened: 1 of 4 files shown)


roles/znc/tasks/main.yml
```diff
@@ -3,28 +3,30 @@
...
-- name: ZNC
+- name: Create ZNC Docker Container
   docker_container:
     name: znc
     image: znc
     pull: true
     volumes:
       - "{{ znc_data_directory }}:/znc-data:rw"
       - "/etc/timezone:/etc/timezone:ro"
+    network_mode: "bridge"
```
davestephens, line 18: Not clear why this was added?

PurpleNinja225: To give confirmation that the container is run on the bridge mode network as apposed to the host network. It also makes it easy for someone to change a containers network mode if they wanted to use a different method, like overlay or macvlan. It's also on all of the other PRs I've moved over to roles and on a few others, such as openhab.

davestephens: OK, I know this is picky but I'm going to remove these as I merge the PRs (which I am very grateful for!). 

Few reasons for that:
- This is a change to what's currently being done at the moment.  
- Adding options that are set to defaults increases the amount of stuff required to be understood. Docker's default is bridge networking, so people shouldn't be surprised by the network mode. 
- If we really think people need to change it, then it should be set as a parameter people can set through role variables

PurpleNinja225: I can add that functionality

PR #422, opened 2021-02-16
Overview: mention Vagrantfile in the Experiment section

axelsimon:
I figured it was worth mentioning the Vagrant option for testing out Ansible-NAS in the overview, and linking to the testing page.

Also mentioning libvirt for good measure, as not everyone uses Virtualbox.


davestephens:
Good points, thank you 👍

PR #423, opened 2021-02-19
add ddns_updater application

d-rk:
I don't use cloudflare but a different provider, so I replaced cloudflare_ddns role with a ddns_updater role.
The role uses https://github.com/qdm12/ddns-updater which supports lots of providers and is easy to use. 
Of course it also supports cloudflare.

**Any other useful info**:
This PR is not complete, yet. The role is working, but I need some opinion on how to continue with this. Afterwards I can finish it up.
So here are things I'm not sure how I should implement it:
- Currently I just replaced cloudflare and the original variables are just dropped. Should there be an automatic migration? Should I write a template that generates a config based on the original cloudflare variables?
- I included an example how the config for cloudflare should look like. Since I have no cloudflare account I would be glad if someone could test that.

### DIFF
--- a/roles/ddns_updater/defaults/main.yml
+++ b/roles/ddns_updater/defaults/main.yml
@@ -1,0 +1,42 @@
+---
+# enable the role
+ddns_updater_enabled: false
+
+# Data directory for config file
+ddns_updater_data_directory: "{{ docker_home }}/ddns_updater"
+
+# enable in order to make web ui reachable from external
+ddns_updater_available_externally: "false"
+
+# Port for the web ui
+ddns_updater_port: 8202
+ddns_updater_hostname: "ddns"
+
+# User to run the container with
+ddns_updater_user_id: "0"
+
+# Token to be used to send messages via gotify
+# ddns_updater_gotify_token: "ABCDEFGHIJKLMOP"
+
+# url of gotify server
+gotify_url: "gotify.{{ ansible_nas_domain }}"
+
+# dict to pass in additional environment variables
+# available in the docker images (https://github.com/qdm12/ddns-updater#environment-variables)
+ddns_update_extra_env: {}
+
+# ddns_updater_config describes the content of the config.json
+# used to configure the updater
+# see https://github.com/qdm12/ddns-updater#configuration
+#
+# using cloudflare should work with a config like this:
+#
+#ddns_updater_config:
+#  settings:
+#    - provider: "cloudflare"
+#      zone_identifier: "{{ cloudflare_zone }}",
+#      domain: "{{ cloudflare_host }}"
+#      host: "@"
+#      ttl: 1
+#      token: "{{ cloudflare_token }}"
+#      proxied: { { cloudflare_proxy | bool } }
--- a/roles/ddns_updater/defaults/main.yml
+++ b/roles/ddns_updater/defaults/main.yml
@@ -2,6 +2,9 @@
 # enable the role
 ddns_updater_enabled: false
 
+# docker image used
+ddns_updater_docker_image:  qmcgaw/ddns-updater:latest
+
 # Data directory for config file
 ddns_updater_data_directory: "{{ docker_home }}/ddns_updater"
 
--- a/roles/ddns_updater/defaults/main.yml
+++ b/roles/ddns_updater/defaults/main.yml
@@ -1,32 +1,39 @@
 ---
-# enable the role
+# Enable the role
 ddns_updater_enabled: false
 
-# docker image used
+# Docker image used
 ddns_updater_docker_image:  qmcgaw/ddns-updater:latest
 
 # Data directory for config file
 ddns_updater_data_directory: "{{ docker_home }}/ddns_updater"
 
-# enable in order to make web ui reachable from external
+# Enable in order to make web ui reachable from external
 ddns_updater_available_externally: "false"
 
 # Port for the web ui
 ddns_updater_port: 8202
+
+# Hostname for subdomain
 ddns_updater_hostname: "ddns"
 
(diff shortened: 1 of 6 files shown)


davestephens:
It would be great to get this added!

Rather than adjust an existing role (and have to engineer an upgrade path for users) it would be simpler to this in its own role. That way people can switch at their own leisure.

Regarding config, users can provide the config block in a dict in their nas.yml, which can then be output using `to_nice_json` straight into the JSON config file...

Thoughts?

random-dwi:
> Regarding config, users can provide the config block in a dict in their nas.yml, which can then be output using to_nice_json straight into the JSON config file...

That's how it already works.

I like the idea of circumventing the migration problem by keeping cloudflare ddns.

I will change that and add the missing documentation.

davestephens:
This looks a bit half-baked - are you planning to create a new PR or fix this one up?

random-dwi:
@davestephens I updated the PR so that it just adds the ddns_updater and made it as configurable as possible.

roles/ddns_updater/defaults/main.yml
```diff
@@ -0,0 +1,42 @@
...
+ddns_updater_hostname: "ddns"
+
+# User to run the container with
+ddns_updater_user_id: "0"
+
+# Token to be used to send messages via gotify
+# ddns_updater_gotify_token: "ABCDEFGHIJKLMOP"
+
+# url of gotify server
+gotify_url: "gotify.{{ ansible_nas_domain }}"
```
random-dwi, line 22: in order to avoid having to define `gotify_url` for every application that wants to send to gotify we probably have to define this variable in all.yml as well

roles/ddns_updater/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+  set_fact:
+    ddns_update_env: |
+      TZ: "{{ ansible_nas_timezone }}"
+      {% if ddns_updater_gotify_token is defined %}
+      GOTIFY_TOKEN: {{ ddns_updater_gotify_token }}
+      GOTIFY_URL: {{ gotify_url }}
+      {% endif %}
+      {% for env in ddns_update_extra_env|dict2items %}
+      {{ env.key }}: {{ env.value }}
+      {% endfor %}
```
random-dwi, line 26: I put some thought into how to make it configurable. Perhaps thats also an option for other roles as well?
With this one could also exchange the docker image and use one with slightly different env variables without need to modify the role itself.

bcurran3:
I've been using this in my AN setup since last year and have worked with the developer (Quentin is GREAT guy!). 

This is a good addition to AN.

roles/ddns_updater/defaults/main.yml
```diff
@@ -0,0 +1,45 @@
...
+ddns_updater_extra_env: {}
+
+# ddns_updater_config describes the content of the config.json
+# used to configure the updater
+# see https://github.com/qdm12/ddns-updater#configuration
+#
+# using cloudflare should work with a config like this:
+#
+#ddns_updater_config:
+#  settings:
```
davestephens, line 38: Seeing how this works, I think it's a little over-engineered. We're expecting users to mentally map a json config example in the container docs to YAML, which then gets mapped back to JSON and written out to a template file when the role is run. It would be easier and less brain-hurty to simply accept a json config block and write that out to the config file:

```
ddns_updater_config: >
{
  "settings": [
    {
      "provider": "cloudflare",
      "zone_identifier": "some id",
      "domain": "domain.com",
      "host": "@",
      "ttl": 600,
      "token": "yourtoken",
      "ip_version": "ipv4"
    }
  ]
}

```

roles/ddns_updater/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+
+- name: Generate config.json
+  copy:
+    content: "{{ ddns_updater_config | to_nice_json }}"
+    dest: "{{ ddns_updater_data_directory }}/config.json"
+
+- name: Generate dns docker environment
+  set_fact:
+    ddns_update_env: |
+      TZ: "{{ ansible_nas_timezone }}"
```
davestephens, line 19: `TZ` should be in the task file as its required. it shouldn't be configurable

roles/ddns_updater/tasks/main.yml
davestephens: I think the gotify section is enough here, but could it also be moved straight to the `docker_container` block within the task to save on the extra config complication? If you feel there are other config options that should be defined, include them too and simply create a variable, rather than trying to be overly clever with mapping keys and values to env vars.

davestephens:
I've added some thoughts in comments - you also have linting issues that need fixing! :-)

roles/ddns_updater/defaults/main.yml
random-dwi: of course, thats way better. I updated that part

roles/ddns_updater/tasks/main.yml
random-dwi: I put everything back directly to env of the container

roles/ddns_updater/tasks/main.yml
random-dwi: I get what you mean. I tried to simplify the task and put everything directly in the env of the container.
the only downside is that the container will now try to contact the gotify server with the invalid token from the default and fail, but thats only an error log in the container an is probably ok.

davestephens:
Very late, but thank you for this contribution. :+1: I just borrowed some of the code from your PR to integrate this into Ansible-NAS with the latest role standards, it was a bit too much of a pain to rebase and merge. Hope that's ok!

PR #424, opened 2021-02-19
add NZBHydra 2

allthestairs:
Adds a NZBHydra 2 container. Works very well with the already-included Sonarr/Radarr/NzbGet/etc setup. See #426 

The nas.yml diff looks like I'm deleting the Radarr role but for some reason it is duplicated in the current version. If it needs to be duplicated for some reason I can leave it.

**Any other useful info**:

I don't know how to make git not see some change to the end of `docs/configuration/application_ports.md`. I didn't touch the end.


roles/nzbhydra2/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
...
+
+- name: NZBHydra2
+  docker_container:
+    name: nzbhydra2
+    image: linuxserver/nzbhydra2
+    pull: true
+    volumes:
+      - "{{ nzbhydra2_data_directory }}:/config:rw"
+    ports:
+      - "5076:5076"
```
davestephens, line 17: `nzbhydra2_port` should be used here

roles/nzbhydra2/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
...
+    env:
+      TZ: "{{ ansible_nas_timezone }}"
+      PUID: "{{ nzbhydra2_user_id }}"
+      PGID: "{{ nzbhydra2_group_id }}"
+    restart_policy: unless-stopped
+    memory: 1g
```
davestephens, line 23: Can you make this configurable please, similar to in other roles?

allthestairs: I made both the changes. The first one I just missed but this one was because the role I chose to copy still had hardcoded memory. Would it be useful to make a PR to unify the memory config on all of the roles?

davestephens:
You still have linting issues that need fixing. Look at the "files changed" tab of this PR and you can see where, or you can run yamllint / ansible-lint locally.

Issue #425, opened 2021-02-19
Enable forward authentication for OAuth integration with Traefik


allthestairs:
**Is your feature request related to a problem? Please describe.**
The applications included in ansible-nas have varying degrees of security and hardening. We have a setup to enable any of them to be reverse proxied by Traefik, but not all are well-suited to be exposed and require manual setup of an authentication method for each in potentially different ways.

**Describe the solution you'd like**
Traefik allows for forward authentication middleware to require authentication before particular backends (old terminology) or routers can be accessed. This can be used to protect any Traefik-proxied application using OAuth authentication when combined with a minimal authentication server like [traefik-forward-auth](https://github.com/thomseddon/traefik-forward-auth). This would potentially enable single-sign-on for authentication on all ansible-nas applications proxied by Traefik.

**Describe alternatives you've considered**
Just setting passwords on everything manually I guess. You can also use exclusively VPN access but that is a sort of orthogonal solution and isn't integrated (reasonably so) with ansible-nas anyway.

**Additional context**
I have a [functional branch](https://github.com/allthestairs/ansible-nas/tree/forwardauth) that works. This adds a new container to the traefik ansible tasks to handle forward authentication as well as a variety of variables to enable configuration of (currently Google-only) OAuth on any application.

Applying forward authentication to a particular app with this setup requires one additional Docker label to be applied:

`traefik.http.routers.app.middlewares: "traefik-forward-auth"` 

My current solution is to add one additional variable analogous to `app_available_externally` named `app_forwardauth_enabled`. I then conditionally apply this middleware in ansible using something like this:

`traefik.http.routers.app.middlewares: "{{ app_forwardauth_enabled | bool | ternary('traefik-forward-auth', '') }}"`

I am very much not an ansible expert, so I am not sure if there is a better way to handle this than just creating a new variable for every existing application and adding the conditional label line to all the tasks. Additionally, I am not sure whether it would be a good idea or not to default applications to be authenticated if you enable forward authentication for Traefik.

allthestairs:
To elaborate on the good and bad of using this:

Good:
- This kind of setup can, with some simple config, let you protect all of your exposed apps with OAuth authentication. 
- Gives you a solid additional layer of protection based on software with a good security record if you choose to expose any applications. 
[...]

PR #427, opened 2021-02-19
WIP: Enable Traefik forward authentication and dynamic file configuration

allthestairs:

Adds forward authentication to Traefik to enable OAuth authentication for any ansible-nas integrated application proxied by traefik.

#425 

**Any other useful info**:

Right now this only adds functional authentication middleware to traefik, it doesn't enable authentication for any applications. Right now the best way I can think of to add authentication is the addition of a label and one boolean variable to every ansible task. You _can_ protect an entire traefik entrypoint without touching the individual tasks but there are good reasons not to OAuth every ansible-nas application.

Note: the client id, client secret, and cookie secret in this are examples and do not correspond to any real Google OAuth setup. 

This also includes an unnecessary change to enable file providers in traefik in case anyone wants to manually add some non-ansible application to their traefik setup by dropping some yaml/toml files into a host folder.

### DIFF
--- a/roles/traefik/defaults/main.yml
+++ b/roles/traefik/defaults/main.yml
@@ -27,3 +27,23 @@ traefik_environment_variables:
 # the "{{ docker_home }}/traefik" after changing this setting.
 # traefik_acme_server: "https://acme-staging-v02.api.letsencrypt.org/directory" # staging
 traefik_acme_server: "https://acme-v02.api.letsencrypt.org/directory" # production
+
+# forwardauth using Google OAuth requries a client id and secret
+traefik_forwardauth_hostname: "auth"
+traefik_google_client_id: "28923428222-3jra876g8a7fg3h9ajf83jaf376hafa.apps.googleusercontent.com"
+traefik_google_client_secret: "f3a3-hH3JDe-ganlIUNAnffoe"
+
+# secret must be random: "openssl rand -hex 16"
+traefik_forwardauth_secret: "oh9dga4s8b653d86n854nusid6bnkljs"
+
+# Whitelist of account names to allow
+traefik_forwardauth_whitelist: ""
+
+traefik_forwardauth_environment_variables:
+  PROVIDERS_GOOGLE_CLIENT_ID: "{{ traefik_google_client_id }}"
+  PROVIDERS_GOOGLE_CLIENT_SECRET: "{{ traefik_google_client_secret }}"
+  SECRET: "{{ traefik_forwardauth_secret }}"
+  INSECURE_COOKIE: "true"
+  AUTH_HOST: "{{ traefik_forwardauth_hostname }}.{{ ansible_nas_domain }}"
+  COOKIE_DOMAIN: "{{ ansible_nas_domain }}"
+  WHITELIST: "{{ traefik_forwardauth_whitelist }}"
--- a/roles/traefik/tasks/main.yml
+++ b/roles/traefik/tasks/main.yml
@@ -5,6 +5,7 @@
     state: directory
   with_items:
     - "{{ traefik_data_directory }}"
+    - "{{ traefik_data_directory }}/config"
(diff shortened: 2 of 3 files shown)


davestephens:
I like the idea of this as an option, but how would someone optionally enable this per container without hacking up the individual docker_container tasks within the roles?

allthestairs:
> I like the idea of this as an option, but how would someone optionally enable this per container without hacking up the individual docker_container tasks within the roles?
> 
> I think as a POC this is great, but it needs some thought about how to integrate it without breaking all of the already established patterns in Ansible-NAS.

I put more info about this in the connected Issue. The easiest way to attach the middleware is to add a label to the container, but you can only attach labels at build time. To match what I see as the current patterns as I see them we would need to modify the existing roles to some degree by adding one label line and one boolean configurable to enable. That way no one needs to touch the roles directly, and could just set `app_authentication = "true"` in their config and it would enable it. 

That is definitely the "simplest" way to integrate this though it does require modifying every role and having it in whatever template someone might use to add more roles in the future the same way the existing Traefik config is. I don't know of any ways to attach a label to a container in ansible without making any change to the docker_container directive at all but maybe it is possible, I don't know too much advanced ansible. 

You can also use a file-based "dynamic configuration" for Traefik. Doing it this way is feasible, but this would require a mapping of config variable names to containers which I haven't seen but I haven't dug through all of anisble-nas. Hypothetically, we could use ansible to generate a file based on searching for specific config variables, somehow discovering which container router needs to be given the middleware, and adding it to the yaml/toml file. The more work you put on the user in the config the easier this is. Making `app_authentication = "true"` is hard unless all of the variables always match our Traefik router names. Making `app_authentication = "true"`, `app_auth_router = "traefik.http.routers.app"` work would be easy but then users would have to look up what the traefik router name is to enable it in the config or (again) we would need to add that variable to every role.

dtee:
I am bit late to the party, but I ended up doing something similar to secure apps that do not auth by default like dashy and code-server. Here's  how I ended enabling per app . 

Modified main.yml and added the middle ware for apps I want protected.
```
          traefik.http.routers.code_server.middlewares: "traefik-forward-auth"
```

I am bit too lazy to add a proper task in ansible-nas, but here's what the config looks like in docker-compose:

```
  traefik-forward-auth:
    image: thomseddon/traefik-forward-auth:2.1.0
    deploy:
      restart_policy:
        condition: unless-stopped
    ports:
      - "4181:4181"
    environment:
      - PROVIDERS_GOOGLE_CLIENT_ID=
      - PROVIDERS_GOOGLE_CLIENT_SECRET=
      - SECRET=something-random
      # INSECURE_COOKIE is required if not using a https entrypoint
      - LOG_LEVEL=debug
      # comment out AUTH_HOST if you'd rather use individual redirect_uris (slightly less complicated but more work)
      - AUTH_HOST=
      - COOKIE_DOMAIN=
      - WHITELIST=
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.auth.rule=Host(``)"
      - "traefik.http.routers.auth.tls.certresolver=letsencrypt"
      - "traefik.http.routers.auth.entrypoints=websecure"
      - "traefik.http.routers.auth.tls=true"
      - "traefik.http.routers.auth.tls.domains[0].main="
      - "traefik.http.routers.auth.tls.domains[0].sans="     
      - "traefik.http.services.auth.loadbalancer.server.port=80"
      
      - "traefik.http.routers.auth.service=auth@docker"
      - "traefik.http.routers.auth.middlewares=traefik-forward-auth"
      - "traefik.http.middlewares.traefik-forward-auth.forwardauth.address"
      - "traefik.http.middlewares.traefik-forward-auth.forwardauth.trustForwardHeader=true"
      - "traefik.http.middlewares.traefik-forward-auth.forwardauth.authResponseHeaders=X-Forwarded-User"
```

Issue #400, opened 2021-02-02
Documentation for roles?


begna112:
Applications appear to be moving from tasks to roles, out of the nas.yml in the base directory, and into the roles folder. Currently, the documentation doesn't present many(any?) of the roles or how/where to override them. 

For example, Plex is missing from the Applications list. I believe that's because it's now a role. I was originally just copying and overriding settings from the base nas.yml to enable them. But Plex isn't in there at all so I couldn't figure out how to enable it. 

Beyond just enabling it, how do we override settings? 

In my Plex example, what if I don't want a Photos volume, but I want to add one for Music Videos or some other media type? What is recommended method of modifying the new roles while still being able to update ansible-nas? Presumably the method of updating should be a `git pull`, but we don't want to overwrite our changes every time we pull. 

Some clarity and updated documentation around this would be really appreciated, especially for someone who is using ansible for the first time with this.

For documentation, I'm referring to here: https://davestephens.uk/ansible-nas/

If there is some other more up to date documentation, please let me know. 

Thanks!

PurpleNinja225:
Documenation Definitiely needs updating, I there is a bit of a hold on it until all the apps are moved over to rolls, but if @davestephens can shoot me an outline of how he wants new documentation I can take care of doing most of the heavy lifting.

georgejung:
Hi there, I am happy to help too. I did a fork, but hesitated to go all the way through with a pull request since I'm not sure I completely know what I am doing. :) I didn't have to change too much but I have recently (December - present) done a few fresh installs on an old box running ubuntu 18 and now 20 and have noticed some issues.

Correct me if I'm wrong but the vast majoirty of the settings from the previous all.yml are still valid, they've just been removed from all.yml so you need the historical version to add them back into nas.yml?

I too would be curious how it works re changing the roles and upgrading. or if we can add volumes from nas.yml?

davestephens:
Dump all your personal changes into your inventory `nas.yml` - there is a doc page on it here: https://davidstephens.uk/ansible-nas/upgrading/

PR #428, opened 2021-02-19
Minor update of installation instructions for roles

allthestairs:

Just a tiny change to the installation instructions to make it more obvious where to find settings for applications switched to roles.

This is just a band-aid on #400 that at least points new people in the right direction to find settings for all the role-migrated applications.

**Any other useful info**:

I'm sure there is plenty of other documentation changes that would be useful but this would probably save brand new people some hassle.


davestephens:
Perfect 👍

PR #433, opened 2021-02-28
Add dokuwiki

eniad:
**What this PR does / why we need it**:

Add [DokuWiki](https://www.dokuwiki.org/), an open source wiki that doesn't require a database

PR #434, opened 2021-02-28
add dokuwiki

eniad:

Add DokuWiki, an Open Source wiki that doesn't require a database

**Any other useful info**:

Simple to use. Allows edits of content files directly. Used as a personal wiki for notes.

---

Squashed version of https://github.com/davestephens/ansible-nas/pull/433


davestephens:
This is awesome!

Please could I request that you flip it to the new "role" format?

The hard work you've done is the same, it just properly contains the change from the rest of the playbook by a better folder structure. (look under "roles" for examples. 

And sorry - I need to update the docs, so this is largely my fault!

eniad:
Will do. I made this change (and a few others) before you switched over. I'll use the new structure. It is much cleaner.

roles/dokuwiki/tasks/main.yml
```diff
@@ -0,0 +1,26 @@
...
+      - "{{ dokuwiki_port }}:80"
+    restart_policy: unless-stopped
+    memory: 1g
+    labels:
+      traefik.enable: "{{ dokuwiki_available_externally }}"
+      traefik.http.routers.dokuwiki.rule: "Host(`{{ dokuwiki_hostname }}.{{ ansible_nas_domain }}`)"
+      traefik.http.routers.dokuwiki.tls.certresolver: "letsencrypt"
+      traefik.http.routers.dokuwiki.tls.domains[0].main: "{{ ansible_nas_domain }}"
+      traefik.http.routers.dokuwiki.tls.domains[0].sans: "*.{{ ansible_nas_domain }}"
+      traefik.http.services.dokuwiki.loadbalancer.server.port: "{{ dokuwiki_port }}"
```
davestephens, line 26: This should be set to the port the container publishes on (80), not the mapped port

roles/dokuwiki/tasks/main.yml
```diff
@@ -0,0 +1,26 @@
...
+  docker_container:
+    name: dokuwiki
+    image: linuxserver/dokuwiki:latest
+    pull: true
+    volumes:
+      - "{{ dokuwiki_data_directory }}:/config/dokuwiki/data:rw"
+    ports:
+      - "{{ dokuwiki_port }}:80"
+    restart_policy: unless-stopped
+    memory: 1g
```
davestephens, line 19: Please can you make this configurable as with other tples

roles/dokuwiki/tasks/main.yml
```diff
@@ -0,0 +1,26 @@
...
+  file:
+    path: "{{ item }}"
+    state: directory
+  with_items:
+    - "{{ dokuwiki_data_directory }}/data"
+
+- name: DokuWiki Docker Container
+  docker_container:
+    name: dokuwiki
+    image: linuxserver/dokuwiki:latest
```
davestephens, line 12: Most linuxserver containers accept uid/gid, do we need this here to permission the data correctly?

eniad: I set the uid/gid to 1000:1000. It works. I wasn't sure about those, so I checked https://hub.docker.com/r/linuxserver/dokuwiki. I understand those Docker permissions better now.

roles/dokuwiki/tasks/main.yml
eniad: Missed that one. Fixed now.

roles/dokuwiki/tasks/main.yml
eniad: Done. Which person resolves PR conversations?

davestephens:
Awesome, thank you! 👍

Issue #435, opened 2021-03-01
With the fix for #253 watchtower_memory needs to be configured; default is broken


delgurth:
While fixing https://github.com/davestephens/ansible-nas/issues/253, using the following commit the memory is configurable, but the default value for watchtower is not set:

https://github.com/davestephens/ansible-nas/commit/a645bd809feecb0bcada1c6344c5d964e0704759#diff-cbd16e0c3e395a05570307f304208e655255ee79bbe657241200e19e5d86defeR20

I'll make a pull request to fix this.

davestephens:
Oops, good spot.

PR #436, opened 2021-03-01
Make the watchtower_memory default an actual default

delgurth:
The watchtower_memory default was not set. Fixes https://github.com/davestephens/ansible-nas/issues/435


davestephens:
Thanks 👍 I thought the 50-odd changes I had made went a bit too well ;-)

iddinkwessel:
It was the last change of them all. Just to be sure I checked the entire commit, saw no breaking changes. And it was ansbile-playbook who spotted the issue, not me.

jackett_memory was the only one also missing a commit for the default. But seems that default was already configured in a previous commit.

Issue #437, opened 2021-03-02
Grafana does not register in traefik correctly


delgurth:
It seems that the grafana route in traefik has not been tested properly since the switch to traefik 2. At least  I assume this is the reason it no longer works for me. I'll make a pull request to fix it.

Current error message from traefik:

time="2021-03-02T20:18:39Z" level=error msg="field not found, node: grafana" providerName=docker container=grafana-9a84f86d61fc396680d15a2ff31b5e48901fc50e60487982251b2dd7230bfabb

codebrewer:
You beat me to submitting a PR!

Did you also notice that the `stats.yml` config no longer works correctly with InfluxDB 2.x? There are a number of changes between InfluxDB 1.x and 2.x, the two most noticeable being that data is now stored by default at `/var/lib/influxdb2` rather than `/var/linux/influxdb` and that clients must now authenticate to post data (I see Traefik log 401 errors from Telegraf and get no data showing in Grafana).

InfluxDB has support for migrating data from 1.x to 2.x (see [here](https://hub.docker.com/_/influxdb?tab=description)) but I'm assuming that one reason this Grafana bug has been around for a while might be that Netdata is now favoured by Ansible NAS users? If that's the case then pinning the `influxdb` image at tag `1.8` might be the way to go (it would at least restore a working setup while updating the config to use the migration feature is considered/implemented).

delgurth:
@codebrewer I just upgraded from a 1 year old git checkout and working from issue to issue. Got the traefik stuff working now (although there seems to be a concurrency issue there when getting a new certificate, yet another thing to debug but lower prio since it works now). Haven't looked at grafana, just got it working in traefik.

Also looking into a way to actually test the ansible-nas playbooks. Since the docker containers running does not say that it's actually working. I first experienced that last year when I added the gitlab container. It didn't function properly because the default 4Gb memory was only enough in case you also had 4Gb of swap space. My box didn't have any swap space, so gitlab didn't start.

codebrewer:
@delgurth perhaps you'll be able to confirm the InfluxDB problem soon too then :smile:

@davestephens it would be a shame to leave Grafana broken so what's your opinion on how to fix the InfluxDB part of the problem? Pin at 1.8? Update the config to migrate existing users from 1.x to 2.x?

I think I'll use Netdata myself though (I'm mesmerised by all its pretty graphs...).

davestephens:
[...]

PR #438, opened 2021-03-02
Grafana was not linked in traefik anymore

delgurth:
This pull request fixes https://github.com/davestephens/ansible-nas/issues/437. Guess the stats should be migrated to a role as well, but didn't want to mix fixing a bug and changing from task to a role in the same commit.


davestephens:
Good spot, thank you! 👍

PR #443, opened 2021-03-08
change jackett tag to just jackett

eniad:

Changed tag names for roles change `transmission_with_openvpn` and `jackett`

Removed "false" from `jackett` tag so it is just "jackett". It was previously "jackettfalse", which looked like a typo.

Removed "_enabled" from `transmission_with_openvpn` tag for the same reason.

### DIFF
--- a/nas.yml
+++ b/nas.yml
@@ -250,7 +250,7 @@
 
     - role: transmission-with-openvpn
       tags:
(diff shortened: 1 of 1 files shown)


davestephens:
👍 good spot, thanks!

PR #444, opened 2021-03-10
Add Piwigo

rahul-r:
Add Piwigo role to ansible-nas. Piwigo is a web based photo management application.

None
Fixes #

**Any other useful info**:

### DIFF
--- a/docs/applications/piwigo.md
+++ b/docs/applications/piwigo.md
@@ -1,4 +1,4 @@
-# pyLoad
+# Piwigo
 
 Homepage: [https://www.piwigo.org/](https://www.piwigo.org/)
 
--- a/docs/applications/piwigo.md
+++ b/docs/applications/piwigo.md
@@ -1,0 +1,11 @@
+# pyLoad
+
+Homepage: [https://www.piwigo.org/](https://www.piwigo.org/)
+
+Piwigo is open source photo gallery software for the web. Designed for organisations, teams and individuals
+.
+## Usage
+
+Set `piwigo_enabled: true` in your `inventories/<your_inventory>/nas.yml` file.
+
+piwigo's web interface can be found at http://ansible_nas_host_or_ip:8033
--- a/roles/piwigo/defaults/main.yml
+++ b/roles/piwigo/defaults/main.yml
@@ -1,0 +1,19 @@
+---
+piwigo_enabled: false
+piwigo_available_externally: "false"
+
+# directories
+piwigo_data_directory: "{{ docker_home }}/piwigo"
+
+# uid / gid
+piwigo_user_id: "1000"
+piwigo_group_id: "1000"
+
+# network
+piwigo_port: "8033"
+piwigo_hostname: "piwigo"
+
+# username / passwords
+piwigo_sql_user: "piwigo-user"
+piwigo_sql_password: "piwigo-pass"
+piwigo_sql_root_password: "piwigo-secret"
--- a/roles/piwigo/tasks/main.yml
+++ b/roles/piwigo/tasks/main.yml
@@ -1,0 +1,40 @@
+---
+- name: Create Piwigo directories
+  file:
+    path: "{{ item }}"
+    state: directory
+  with_items:
+    - "{{ piwigo_data_directory }}/config"
+
+- name: Piwigo Mysql Docker Container
+  docker_container:
+    name: piwigo-mysql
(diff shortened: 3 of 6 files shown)


Maltazar:
I would love this to :)

roles/piwigo/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    restart_policy: unless-stopped
+
+- name: Piwigo Docker Container
+  docker_container:
+    name: piwigo
+    image: ghcr.io/linuxserver/piwigo
+    pull: true
+    volumes:
+      - "{{ piwigo_data_directory }}/config:/config:rw"
+      - "{{ piwigo_data_directory }}/pictures:/pictures"
```
random-dwi, line 30: Photos are usually placed under `{{ photos_root }}`.
I would assume this should at least be the default location to mount here

georgejung (changes requested):
I have not tested this yet, but shouldn't the nas.yml file have a section for piwigo?
somehting along the lines of:
  - role: piwigo
      tags:
        - piwigo
      when: (piwigo_enabled | default(False))

I'll try adding it to mine to confirm it works well. I would really like to see this added as its a key feature for NAS, and piwigo has great app support across platforms (even if android is limited to 100 pics in an album).

One comment would be, is it possible to re-use an existing mySQL dbinstance? or is that frowned upon? I can see the advantages of keeping things seperate, but i would think it should be quite a bit more efficient to run one instance with multiple databases vs mutiple DB instances...

Thanks for this

roles/piwigo/defaults/main.yml
```diff
@@ -0,0 +1,19 @@
+---
+piwigo_enabled: false
+piwigo_available_externally: "false"
+
+# directories
+piwigo_data_directory: "{{ docker_home }}/piwigo"
```
georgejung, line 6: include the following line in # directories:
piwigo_photos_directory: "{{ photos_root }}"

roles/piwigo/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    ports:
+      - "{{ piwigo_port }}:8033"
```
georgejung, line 32: here the reference to piwigo port is good but after the : it should be 80, not 8033

roles/piwigo/tasks/main.yml
georgejung, line 30: I used this volume mounting for piwigo docker container. It allowed the sync to pull directly from my samba photo share.
      - "{{ piwigo_photos_directory }}:/config/www/gallery/galleries/photos"

docs/applications/piwigo.md
```diff
@@ -0,0 +1,11 @@
...
+
+Homepage: [https://www.piwigo.org/](https://www.piwigo.org/)
+
+Piwigo is open source photo gallery software for the web. Designed for organisations, teams and individuals
+.
+## Usage
+
+Set `piwigo_enabled: true` in your `inventories/<your_inventory>/nas.yml` file.
+
+piwigo's web interface can be found at http://ansible_nas_host_or_ip:8033
```
georgejung, line 11: worthwhile noting that when doing the webui install the localhost will be the ip of the mysql docker container. localhost didn't work for me. Previously i used "db" with 'services' defined in my docker compose and that worked but services are not defined in the ansible tasks yml and I don't know how to do that in ansible.

roles/piwigo/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    restart_policy: unless-stopped
+    labels:
+      traefik.enable: "{{ piwigo_available_externally }}"
+      traefik.http.routers.piwigo.rule: "Host(`{{ piwigo_hostname }}.{{ ansible_nas_domain }}`)"
+      traefik.http.routers.piwigo.tls.certresolver: "letsencrypt"
+      traefik.http.routers.piwigo.tls.domains[0].main: "{{ ansible_nas_domain }}"
+      traefik.http.routers.piwigo.tls.domains[0].sans: "*.{{ ansible_nas_domain }}"
+      traefik.http.services.piwigo.loadbalancer.server.port: "8033"
```
georgejung, line 40: I added the following at the bottom, in line with labels, ports and restart_policy, so that at the install screen i can use db as host.
links:
  - piwigo-mysql:db

PR #446, opened 2021-03-12
change docker image for airsonic role

random-dwi:
This PR makes the airsonic docker image configurable.
Additionally, the default image is changed to https://hub.docker.com/r/linuxserver/airsonic because it also supports arm architectures.


davestephens:
Arm architectures are not something that I've said is supported, so I'm not keen on merging this without a more holistic approach across the project. I would like to support arm (and I've even hacked Ansible-NAS onto a Pi myself) - but it needs a bit more thought than just changing this one role. 

Also, I'm not sure that allowing changing the whole docker image (rather than just the tag) is desirable, as it then says that any docker image is supported - which we can see from your PR here is not the case due to path and config differences.

random-dwi:
just closing this for now

PR #447, opened 2021-03-14
Clear up confusion with 2 nas.ymls

nough:
There are two files named nas.yml, one in `~/ansible-nas/` and one in ~/ansible-nas/inventories/my-ansible-nas/group-vars/`. This PR adds extra text to line 8 of the installation instructions to clear this up.


davestephens:
Thanks for the idea but I'm not going to merge this. If you follow the instructions you won't end up in the wrong directory, and the instructions don't specifically tell you to clone to your home dir.

PR #452, opened 2021-03-28
Update n8n.md

axelsimon:
Saw a typo, went to correct it, decided i should also learn what n8n is given the page didn't say, here's the result of these two things :)

Improves doc.


davestephens:
👍 thanks!

PR #459, opened 2021-04-17
add OpenNMS

eniad:
What this PR does / why we need it:

Add [OpenNMS](https://www.opennms.com//), an open source enterprise grade network management application platform

Any other useful info:

Using this to monitor and diagnose home network outages. This application might be overkill for ansible-nas. I set it up, so I figured I'd put in a PR to let someone else decide.


eniad:
I'm going to close this. While OpenNMS is useful, there are lighter weight options available.

PR #466, opened 2021-06-13
Scrutiny

### LINKED ISSUE #2: SMART Disk Monitoring
Check disks for SMART alert statuses and alert accordingly.

allthestairs:

It converts the stats task to a role and adds a container running Scrutiny. This container runs a webui with current S.M.A.R.T. data, keeps a history of drive temperatures, and handles notifications using shoutrrr URLs. A NAS obviously benefits heavily from monitoring the hardware that underlies all of its storage. This provides that using what seems to be a reasonably well-maintained container.

Fixes #2 

**Any other useful info**:

This provides template files for both Scrutiny config files and makes the most important two things configurable: which drives to expose to the container for smartd (in the task docker_compose) and the URLs for notifications (in the scrutiny.yaml config template).


davestephens:
Despite the dependency, Scrutiny needs to be separated from the stats change into separate PRs.  

Also, it looks like some prior functionality has been removed (ie copying in the Ansible-NAS dashboard). Presumably this was a mistake rather than a deliberate change!

I appreciate you submitted this months ago, so let me know if you'll pick this up, if not I'll try and tackle it in the next week or so.

allthestairs:
It has been a while. I see the stats task is gone anyway. I cherry picked my changes, moved it to its own role, updated it to current role style, and added documentation. 

Just as a side note, as I was cleaning up my stats role I noticed there is some detritus left over from the deletion of the stats task: grafana and telegraf files and templates that are no longer used.

AnalogJ:
FYI, if you're using the official image (`ghcr.io/analogj/scrutiny`) the correct container path is `/opt/scrutiny/*` (it was changed in v0.4.x)

djosh9855 (approved):

PR #471, opened 2021-08-07
WIP Split big variable file in smaller more managable ones

prk0ghy:
ignore python virtualenv
Update nextcloud

Split the group variables in smaller more manageable files.
Update nextcloud.

Fixes #

**Any other useful info**:


PurpleNinja225:
Can you elaborate a bit more on what you are trying to accomplish by splitting the  all.yml? irrc the all.yml is more for showing off a complete list of variables available/fallback 🤔. Configuration changes should be applied to  `inventories/mynas/group_vars/nas.yml` as they will have priority. See [Controlling how Ansible behaves: precedence rules](https://docs.ansible.com/ansible/latest/reference_appendices/general_precedence.html#controlling-how-ansible-behaves-precedence-rules)
>Precedence categories
>
>Ansible offers four sources for controlling its behavior. In order of precedence from lowest (most easily overridden) to highest (overrides all others), the categories are
>
>       Configuration settings
>
>        Command-line options
>
>        Playbook keywords
>
>        Variables
>
>Each category overrides any information from all lower-precedence categories. For example, a playbook keyword will override any configuration setting.
>
>Within each precedence category, specific rules apply. However, generally speaking, ‘last defined’ wins and overrides any previous definitions.

By setting our variables in `inventories/mynas/group_vars/nas.yml` this allows us to set up all our custom changes to roles that need them and still use the defaults for others if we want to enable them as ansible will still fallback on either the `group_vars/all.yml` or the `roles/defaults/main.yml`

Also updating Nextcloud changes should be a seperate pull request.

PR #478, opened 2021-08-21
Update nas.yml

sparkeh:
Notice a duplicate radarr role within `nas.yml` file. PR to resolve this.

Thanks
Denis


davestephens:
Weird, good spot - thanks!

PR #480, opened 2021-08-30
Fix typo in jellyfin config for books and audiobooks directories

nicpayne713:
Title basically sums it up - fixing a small Jellyfin config typo


davestephens:
Good spot, thanks!

PR #485, opened 2021-09-02
Disable AppArmor via 'unconfined' profile

wmudge:
To stop log spamming by AppArmor. 

**Any other useful info**:
A better long-term solution might be to create (or find) a proper AppArmor profile ala https://docs.docker.com/engine/security/apparmor/.


davestephens:
:+1: thanks!

PR #491, opened 2021-10-19
correcting netbootxyz main.yml variable name error

nickjones33:

netbootxyz_available_externally is the name in defaults, netbootxyz_externally was the name in main, just correcting so that both are netbootxyz_available_externally

**Any other useful info**:

fixes an error thrown during the netbootxyz portion of the playbook where it cannot locate {{ netbootxyz_externally }}


davestephens:
:+1: thanks!

PR #494, opened 2021-11-10
docker: allow customizing daemon.json

khartahk:
Allows customizing `/etc/docker/daemon.json`

**Any other useful info**:
I'm using a different subnet for my docker containers because the default one conflicts with my IPTV providers subnet and I can't use nextpvr or tvheadend with it.

I'm specifying an additional `bip` line in `docker_daemon_json` variable:
```yaml
docker_daemon_json:
- '"data-root": "{{ docker_image_directory }}"'
- '"storage-driver": "{{ docker_storage_driver }}"'
- '"bip": "172.29.0.1/16"'
```


davestephens:
I fixed this in a more "ansible" way in https://github.com/davestephens/ansible-nas/commit/863cdc20fdd16ed0ffdde4bb8d8e47659657caec, thanks for the suggestion :+1:

PR #503, opened 2021-12-21
syncthing: don't run in host network mode, specify uid/gid

khartahk:
- adds ability to specify as which user syncthing should run (files can have owner other then root)
- removes requirement to run syncthing in host network mode.


davestephens:
:+1:

PR #308, opened 2020-05-02
ADDITION: Unifi (controller)

bcurran3:
**What this PR does / why we need it**:

Adds the popular Ubiquiti Unifi Controller to Ansible-NAS.

**Which issue (if any) this PR fixes**:

Fixes # https://github.com/davestephens/ansible-nas/issues/153

**Any other useful info**:

Unfortunately complex - but Ansible-NAS users should be abled enough.

PR #506, opened 2021-12-26
Add UniFi Network Application stack

GeriYatola:

This adds [UniFi Network Application](https://help.ui.com/hc/en-us/articles/360012282453) stack to Ansible-NAS. This is a valuable addition to any owners of [Ubiquiti UniFi products](https://ui.com/consoles), since it completely replaces a separate UniFi Cloud Key or manually deploying the Network Application stack.

**Any other useful info**:

The unifi-docker container is provided by [Jacob Alberty](https://github.com/jacobalberty/unifi-docker), and the stack is based on their Docker Compose example. We define a separate network (`unifi`) for the stack to assert the default ports can always be used.

We run our Ansible-NAS on UTC timezone, but prefer having UniFi controller on the local TZ. That's why the TZ for the UniFi container can be overwritten with `unifi_timezone`

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ Ansible config and a bunch of Docker containers.
 * [Traefik](https://traefik.io/) - Web proxy and SSL certificate manager
 * [Transmission](https://transmissionbt.com/) - BitTorrent client (with OpenVPN if you have a supported VPN provider)
 * [Ubooquity](http://vaemendis.net/ubooquity/) - Book and comic server
+* [UniFi Network Application](https://help.ui.com/hc/en-us/articles/360012282453) - Self-host UniFi Network Application to manage UniFi devices
 * [uTorrent](https://www.utorrent.com/) - The best torrent downloading app for beginners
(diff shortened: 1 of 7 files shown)


davestephens:
Thanks for the contribution!

Few comments:

- Variables seem to be mixed up between `group_vars/all.yml` and `roles/unifi/defaults/all.yml`, some of which are also repeated. Please can you move all the vars into the role.
- Running as the unifi user is optional, but adds a lot of extra complexity (and weirdness on Ubuntu ref systemd-coredump user) - please can you remove the `user:` stanza on the `docker_container` definition which will run the container as root. This also follows the standard of all the other containers in Ansible-NAS. There will be some other cleanup required because of this.

Aside from these minor points - good to merge!

bcurran3:
@GeriYatola - not sure if it's useful at all, but I attempted to get this added about two years ago. It's out of date, but you can look at https://github.com/davestephens/ansible-nas/pull/308 for any possible useful information. It worked fine even though it never got commented on/reviewed/merged. :-( I've since switched to a UDMPRO and don't care any more. :-)

Issue #512, opened 2022-01-09
Firefly III image update


slowesc:
After installation, Firefly page states:
The Firefly III image has moved
For the latest version of Firefly III, use fireflyiii/core:latest
For the develop (unstable) version of Firefly III, use fireflyiii/core:develop

My apologies for the inconvenience.
All the best,
James

davestephens:
Thanks for the heads up - if you want to have a go at updating I'll accept
a PR to fix this. 👍

PR #513, opened 2022-01-09
Update Firefly III image

slowesc:
Firefly III image has moved
Latest version of Firefly III at fireflyiii/core:latest

#512


davestephens:
Thanks! :+1:

PR #525, opened 2022-05-19
default smb loation for audiobooks as audiobook root

nicpayne713:

An unexpected default samba share location for audiobooks given that the rest of the media's default share locations are their defined root directories.


davestephens:
Good spot, thanks!

PR #528, opened 2022-05-24
Add RSS-Bridge

Dylancyclone:
Add [RSS-Bridge](https://rss-bridge.github.io/rss-bridge/).
From [their Github](https://github.com/RSS-Bridge/rss-bridge): 
RSS-Bridge is a PHP project capable of generating RSS and Atom feeds for websites that don't have one. It can be used on webservers or as a stand-alone application in CLI mode.

Important: RSS-Bridge is not a feed reader or feed aggregator, but a tool to generate feeds that are consumed by feed readers and feed aggregators.

N/A

**Any other useful info**:


Dylancyclone:
As soon as this one is merged, I'll rebase the last one

davestephens:
:+1:

PR #539, opened 2022-06-19
Update paperless_ng.md

TimJ0212:
Because the variable for Paperless is wrongly described in the docs.


davestephens:
Good spot. Thanks 👍🏻

PR #541, opened 2022-06-26
Add deCONZ.

brumi1024:

Fixes #

**Any other useful info**:

### DIFF
--- a/roles/deconz/tasks/main.yml
+++ b/roles/deconz/tasks/main.yml
@@ -1,0 +1,49 @@
+---
+- name: Check if the ZigBee device exists
+  find:
+    paths:
+      - "/dev"
+    patterns:
+      - "ttyACM*"
+    file_type:
+      any
+  register: zigbee_dev
+
+- name: DeCONZ Docker Container
+  docker_container:
+    name: deconz
+    image: "deconzcommunity/deconz:stable"
+    pull: yes
+    state: 'started'
+    ports:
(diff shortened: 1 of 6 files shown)


roles/deconz/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+      any
+  register: zigbee_dev
+
+- name: DeCONZ Docker Container
+  docker_container:
+    name: deconz
+    image: "deconzcommunity/deconz:stable"
+    pull: yes
+    state: 'started'
+    ports:
```
davestephens, line 18: This section can be removed due to `network_mode: host`

roles/deconz/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
+---
+- name: Check if the ZigBee device exists
```
davestephens, line 2: It would be good to follow this task up with a check and error if there are no devices found, rather than the `when` line at the bottom of the `docker_container` definition (which would fail silently).

davestephens (changes requested):
This looks good, thanks - couple of minor tweaks then we should be good to merge.

brumi1024:
Closing this one, as I've been testing Home Assistant's own [ZHA integration](https://www.home-assistant.io/integrations/zha/) and it seems to be an easier way to manage Zigbee networks.

PR #542, opened 2022-06-26
Add Overseerr.

brumi1024:

This PR introduces [Overseerr](https://overseerr.dev), which is a nice, modern UI to handle media requests to Plex users. It's an alternative to Ombi.

### DIFF
--- a/roles/overseerr/tasks/main.yml
+++ b/roles/overseerr/tasks/main.yml
@@ -3,7 +3,6 @@
   file:
     path: "{{ item }}"
     state: directory
-    # mode: 0755
   with_items:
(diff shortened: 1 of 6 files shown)


roles/overseerr/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
+---
+- name: Create overseerr Directories
+  file:
+    path: "{{ item }}"
+    state: directory
+    # mode: 0755
```
davestephens, line 6: Please can you remove the commented out text

davestephens:
Awesome, thanks! 👍

PR #545, opened 2022-07-22
:sparkles: added cloudflared service

### LINKED ISSUE #522: Add cloudflared
Useful for creating direct tunnels to couldflare endpoints, allows the user to set up public services while behind CG NAT and similar. Contrarily to cloudflare dynamic dns,  this container would allow putting ansible-nas services on the internet on demand (turn on or off the cloudflared app), since the only point of access would be provided by this container.

This involves some setup first, since you need to:

- get a cloudflare domain purchased
- create a cloudflare tunnel to use from your machine
- add DNS rules to connect your machine to the cloudflare domain using the cloudflare tunnel
- save the configuration data to a known folder, to be mapped inside the container for use

More info for the setup part on [this link](https://developers.cloudflare.com/cloudflare-one/connections/connect-networks/private-net/#before-you-start). The cloudflared container can be used to do the login/auth/tunnel creation steps as well.

My setup currently uses this container, and the `docker-compose` I created in Portainer looks like this

```yaml
version: "2"
services:
  cloudflared:
    image: cloudflare/cloudflared:2022.1.2
    container_name: cloudflared
    environment:
      - TZ=${TZ}
    volumes:
      - /path_to_my_cloudflared_directory:home/nonroot/.cloudflared/
    restart: unless-stopped
    command: tunnel run ${TUNNEL_NAME}
    network_mode: host
 ```

Let me know if you need other info regarding my setup!

### PR DESCRIPTION
**Pull request by HitLuca**
**What this PR does / why we need it**: Adds Cloudflare Argo tunnel support through cloudflared, very useful when a user is behind a Carrier Grade NAT (often the case when using a SIM card for internet connection)

Fixes #522

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ If you have a spare domain name you can configure applications to be accessible 
 * [Booksonic](https://booksonic.org/) - The selfhosted audiobook server
 * [Calibre-web](https://github.com/janeczku/calibre-web) - Provides a clean interface for browsing, reading and downloading eBooks using an existing Calibre database.
 * [Cloud Commander](https://cloudcmd.io/) - A dual panel file manager with integrated web console and text editor
+* [Cloudflared (Argo tunnel)](https://github.com/cloudflare/cloudflared) - Bypass Carrier Grade NAT (CGNAT) restrictions by serving your services through a direct Cloudflare connection
 * [Cloudflare DDNS](https://hub.docker.com/r/joshuaavalon/cloudflare-ddns/) - automatically update Cloudflare with your IP address
 * [CouchPotato](https://couchpota.to/) - for downloading and managing movies
 * [Deluge](https://dev.deluge-torrent.org/) - A lightweight, Free Software, cross-platform BitTorrent client.
--- a/docs/applications/cloudflared.md
+++ b/docs/applications/cloudflared.md
@@ -1,0 +1,27 @@
+# Cloudflared (Argo tunnel)
+
+Homepage: [https://github.com/cloudflare/cloudflared](https://github.com/cloudflare/cloudflared)
+
+Cloudflare Argo tunnel: [https://blog.cloudflare.com/argo-tunnel/](https://blog.cloudflare.com/argo-tunnel/)
+
+Cloudflare: [https://www.cloudflare.com](https://www.cloudflare.com)
+
+This service is very useful when your NAS doesn't have a static IP and it's situated a [Carrier Grade NAT](https://en.wikipedia.org/wiki/Carrier-grade_NAT). With this approach your NAS is connected directly to the Cloudflare servers, which allow public access to your externally available applications.
+
+## Usage
+
+Set `cloudflared_enabled: true` in your `inventories/<your_inventory>/nas.yml` file.
+
+Set `cloudflared_token` to the one you grab from the Cloudflare Zero Trust Dashboard (more below).
+
+## Specific Configuration
+
+Create a cloudflare account if you don't have one. You will need to purchase a domain name registered with Cloudflare, or migrate an existing one on the platform.
+
+On your cloudflare account page, navigate to the Zero Trust dashboard, and create a Tunnel in the Access/Tunnels menu.
+
+After choosing a name for your tunnel, in the environment section choose Docker, and take note of the token in the code section below (the long string after the `--token` command): this will be your `cloudflared_token`.
+
+In the next page (public hostname) you will need to set some values, which we will delete later: set your domain as your cloudflare hostname, Service as `HTTPS` and `localhost`, then save the changes. Edit the Catch-all-rule to be `https://localhost:443`, then delete the entry you made one step before. Your Cloudflare tunnel will now redirect all incoming requests to your Traefik service runnin on port 443, which will route them accordingly.
+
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -20,6 +20,7 @@ If you have a spare domain name you can configure applications to be accessible
 * [Booksonic](https://booksonic.org/) - The selfhosted audiobook server
 * [Calibre-web](https://github.com/janeczku/calibre-web) - Provides a clean interface for browsing, reading and downloading eBooks using an existing Calibre database.
 * [Cloud Commander](https://cloudcmd.io/) - A dual panel file manager with integrated web console and text editor
+* [Cloudflared (Argo tunnel)](https://github.com/cloudflare/cloudflared) - Bypass Carrier Grade NAT (CGNAT) restrictions by serving your services through a direct Cloudflare connection
```
davestephens, line 23: I think this needs tweaking - the name has been changed to "Cloudflare Tunnel", and it'd be better to mention its primary purpose (increasing security with a private VPN tunnel into the Cloudflare network).

docs/applications/cloudflared.md
```diff
@@ -0,0 +1,27 @@
+# Cloudflared (Argo tunnel)
+
+Homepage: [https://github.com/cloudflare/cloudflared](https://github.com/cloudflare/cloudflared)
+
+Cloudflare Argo tunnel: [https://blog.cloudflare.com/argo-tunnel/](https://blog.cloudflare.com/argo-tunnel/)
+
+Cloudflare: [https://www.cloudflare.com](https://www.cloudflare.com)
+
+This service is very useful when your NAS doesn't have a static IP and it's situated a [Carrier Grade NAT](https://en.wikipedia.org/wiki/Carrier-grade_NAT). With this approach your NAS is connected directly to the Cloudflare servers, which allow public access to your externally available applications.
```
davestephens, line 9: NAT circumvention is a useful feature but not the primary purpose

docs/applications/cloudflared.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Specific Configuration
+
+Create a cloudflare account if you don't have one. You will need to purchase a domain name registered with Cloudflare, or migrate an existing one on the platform.
+
+On your cloudflare account page, navigate to the Zero Trust dashboard, and create a Tunnel in the Access/Tunnels menu.
+
+After choosing a name for your tunnel, in the environment section choose Docker, and take note of the token in the code section below (the long string after the `--token` command): this will be your `cloudflared_token`.
+
+In the next page (public hostname) you will need to set some values, which we will delete later: set your domain as your cloudflare hostname, Service as `HTTPS` and `localhost`, then save the changes. Edit the Catch-all-rule to be `https://localhost:443`, then delete the entry you made one step before. Your Cloudflare tunnel will now redirect all incoming requests to your Traefik service runnin on port 443, which will route them accordingly.
```
davestephens, line 25: Typo in "runnin"

roles/cloudflared/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
...
+  docker_container:
+    name: cloudflared
+    image: cloudflare/cloudflared
+    pull: true
+    env:
+      TZ: "{{ ansible_nas_timezone }}"
+    restart_policy: unless-stopped
+    command: "tunnel --no-autoupdate run --token {{ cloudflared_token }}"
+    network_mode: "host"
+    recreate: true
```
davestephens, line 12: Is it necessary to recreate the container on every run?

davestephens (changes requested):
Few tweaks needed then we should be good to merge 👍🏻

roles/cloudflared/tasks/main.yml
HitLuca: Sorry my mistake, it's left behind from my personal setup in which I was setting watchtower labels for autoupdate, which requires recreating the container

README.md
HitLuca: This makes sense, I just used cloudflared to bypass CGNAT but I see how it's useful to show its other capabilities to the user

allthestairs:
I found myself setting this up recently so I thought I'd toss a link to my [cloudflared branch](https://github.com/allthestairs/ansible-nas/tree/cloudflared) in case either of you found any of it useful. The only real differences are in documentation and the way that I set up the container. Instead of setting it up with host networking, I use a bit of Ansible to identify the host IP address in the bridge network, then configure the containers `/etc/hosts` file to access the host using the hostname specified in `ansible_nas_hostname` as opposed to using host networking and specifying `localhost`.

kamilkrz:
Hi @HitLuca This version looks promissing. Right now I am using @allthestairs approach but I would like to make it more automatic and ready for traefik serving me also internal domains.
Problems:
Probably missing step with adding cname widlcard for a tunnel
After this I am almost home
`ERR Request failed error="Unable to reach the origin service. The service may be down or it may not be responding to traffic from cloudflared: x509: certificate is valid for ___redacted___.traefik.default, not localhost" connIndex=1 dest=http://homepage.domain.com/favicon.ico ip=re.da.ct.ed type=http`

I dug a bit and found https://github.com/aeleos/cloudflared/issues/14#issuecomment-1059480745
What I understand is:
Traefik returns its own default cert instead of correct one.

PR #546, opened 2022-07-22
Added raspberry pi support

HitLuca:
I noticed that using a raspberry pi as platform for this project would result in two issues:

- The geerlingguy.docker role won't work as it's not made for arm architecture
- Enabling universe repositories would fail every time

For these reasons I added the `geerlingguy.docker_arm` role to the requirements of this project, and toggled the two versions (along with universe repositories enabling) through a `x86_architecture` flag.

This setup has been tested multiple times on a Raspberry Pi 4 model B, but should work on any ARM controller running Ubuntu/Ubuntu server.

Lemme know if you need more details, sadly I only have one rpi and cannot test it against other models (won't disregard a raspberry pi care package for testing tho)


HitLuca:
I just noticed my formatter (which automatically runs on file save) also putted double quotes on some strings, sorry about that!

davestephens:
Pi support is something I've had on my list to do for ages, I've actually got a pi somewhere that I manually hacked it on to...

However, I wonder if we could use ansible facts rather than expecting a user to set something manually. `ansible_architecture` seems to contain the string we'd need to work with...

HitLuca:
@davestephens much cleaner now, thanks for the tip! I started learning ansible with this project so there's a ton of stuff I'm still missing :)

README.md
```diff
@@ -114,6 +114,8 @@ gained by automating it. Check out the [docs](https://davestephens.github.io/ans
 
 See [Installation](https://davestephens.github.io/ansible-nas/installation/).
 
+Now Ansible NAS can be deployed on a Raspberry Pi (pi 4b model tested)! Make sure to read the installation docs on how to enable ARM architecture support
```
davestephens, line 117: Does arm support still need to be enabled with the latest changes? 🤔

davestephens (changes requested):
I think this is pretty much there now, just a minor doc tweak with regards to arm support vs needing to " enable" it then I think we're good to go!

Side note - wondering if it's worth mentioning that not all applications have arm builds etc etc?

HitLuca:
Shouldn't all applications work without issues since they are virtualized services? I'm pretty sure docker is platform independent but if you have doubts I can do some reading to make sure

davestephens:
Docker isn't platform independent. It requires the image maintainer to support the platform you're pulling the image onto. 

https://docs.docker.com/desktop/multi-arch/

Also, Docker isn't virtualization - it's containerisation of a running application - which is not the same thing :-)

In summary - the change you've made will work for images that have been built for arm64 too. If they haven't, the container will fail to start.

HitLuca:
Sorry I'm used to interact with people that don't know docker and aren't interested in the details, and an easy first explanation that doesn't go too deep is that it virtualizes an application, thank you for the correction. I will make sure to mention that this project can't guarantee that every application will be available if the platform is not the standard x86_64

SolidRhino:
Is here any action needed to merge this change?

in03:
Hey all, I'm running a Pi 4 and keen to get this up and running. Happy to do some testing if necessary? Any further action needed here? :)

HitLuca:
Happy to get more testing done, it should go without issues as I tested on a 4b myself, but you never know :)

davestephens:
Many thanks for this - I've picked parts of this PR in 791ff3e and removed parts that are no longer relevant.

PR #549, opened 2022-07-23
 Switched from booleans enclosed in strings to standard booleans

### LINKED ISSUE #547: Switch from booleans enclosed in strings to standard booleans
There is some inconsistency when looking at the variables used in this project:
- on one hand you have `*_enabled: bool` variables for enabling applications
- on the other hand you have `*_available_externally: "bool"` for allowing applications to be routed through traefik

From my testing switching from `"bool"` to `bool` shouldn't be an issue, as the only change needed is in the traefik label: going from `: {{ *_available_externally }}` to `: {{ *_available_externally | string }}`

HitLuca:
Refactored booleans from "true"/"false" to true/false where needed

Fixes #547

### DIFF
--- a/docs/applications/gitlab.md
+++ b/docs/applications/gitlab.md
@@ -8,7 +8,7 @@ If Gitea isn't powerful enough for you then consider GitLab. It's a much more po
 
 Set `gitlab_enabled: true` in your `inventories/<your_inventory>/nas.yml` file.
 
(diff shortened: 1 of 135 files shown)


davestephens (approved):

davestephens:
Really nice change to tidy things up, thanks 👍

PR #552, opened 2022-07-25
Big docs refactoring

HitLuca:
Took some time (much more than I thought, but it was worth it) to refactor basically every .md file in the project. With the help of the [Markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) and [Gremlins tracker](https://marketplace.visualstudio.com/items?itemName=nhoizey.gremlins) extensions for VSCode, I tried to fix every issue I found.

Below is a (possibly incomplete) list:
- Bare urls, and urls in the form [ url ] ( url ) replaced with < url > 
- missing empty lines between headers, missing end of file newline
- missing top-level headers, incorrect header types
- missing newlines between elements that should go in (of course) new lines
- lists indentations
- wrong double quotes/hyphenes
- end of line spaces

I also removed a duplicate ansible entry in the gitlab task

Finally, maybe, a pain point: I removed the newlines used to replace what the word-wrap would do in the editor, meaning that

```raw
some text
that has been
written like this
as it would be seen
on a screen
```

is now on a single line. This method of writing breaks a bunch of features of most IDEs (syntax highlighting, text coloring, urls recognition etc.), so I hope this change can be kept.

Overall nothing much will change visually, the biggest thing will be in some application pages you will have this

```raw
Homepage: urlProject: url

project description
```

becoming this

```raw
Homepage: url
Project: url

project description
```

All the linting rules I applied are from [this](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md) list


davestephens (changes requested):
This is great! It would be awesome if we can add it to the linting checks that GitHub actions run - do you want to have a go at adding that too?

HitLuca:
Of course, happy to help!

HitLuca:
markdown linting done, seems to work correctly as it spotted a couple of things for me. No gremlins check added, but I sincerely hope people aren't going to start using greek question marks for fun

image

davestephens (approved):

davestephens:
Awesome addition, thanks!

PR #555, opened 2022-07-27
Converted remaining tasks to proper roles

### LINKED ISSUE #554: roles vs tasks?
Not really a bug, but why do we have a `tasks` folder as well as a `roles` folder? The entries I checked in the tasks folder are the same as roles, but with their variables in `group_vars.yml`, which could also be made cleaner by moving the tasks to be roles with their own file for variables definition

HitLuca:
Since Issue #554 mentioned we still have tasks only because there hasn't been anyone willing to do it, I just took some time and finished the refactoring.

Apart from moving tasks to proper roles I also did the following:
- reordered alphabetically the entries in `nas.yml`, since alphabet is subjective looks like :P
- added missing (?!?) entries in `nas.yml`. Basically some applications couldn't be used even if the `*_enabled` variable was set to `true` because it didn't have an entrypoint
 

Fixes #554


davestephens:
This is an amazing addition, thank you! 🙏 I need to sit down for an hour and go through this which I want to do this weekend - GitHub is saying there are merge conflicts, but I'm not seeing what the issue is whilst looking from my mobile.

HitLuca:
I fixed the merging issue, we just had diverging branches since the same file was updated differently by my branch and master

davestephens (approved):
💥

PR #558, opened 2022-07-28
Added postfix gmail relay

HitLuca:
Sets up postfix using a Gmail app password, allowing the NAS to send emails using the `mail` command. Very useful for:
- sending alerts for zpool scrubs outcomes
- sending errors for failed SMART checks on drives
- sending alerts for backup outcomes/failures (e.g. when using Duplicati)

Tested on both x86_64 and arm64 architectures, no issues emerged.

I would plan on adding some documentation on how to set up some of the above usecases in the `zfs_configuration.md` file or similar (since I use this setup for the cases above), let me know :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ If you have a spare domain name you can configure applications to be accessible 
 * [Piwigo](https://piwigo.org/) - Photo Gallery Software
 * [Plex](https://www.plex.tv/) - Plex Media Server
 * [Portainer](https://portainer.io/) - for managing Docker and running custom images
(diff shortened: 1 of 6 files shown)


HitLuca:
Hmm need to check on this later, linting issue but not straightforward to fix from mobile

davestephens:
I like the thought of having Postfix up to send email, but I'm not sure I like how tied in the implementation is to gmail. For example, I have mine configured to use sendgrid. 

I think without too much effort this could be generalised to be a Postfix SMTP relay setup, where someone just configures their own SMTP details via a few variables. 

What do you think?

HitLuca:
Yeah I would need to check other examples, but I suspect that we can adapt it fairly easily to be general. I'll work on it

PR #559, opened 2022-07-30
komga to nasyml

jaimesBooth:
Komga role reference is missing from nas.yml. Restoring this enables Komga to be used. 

Fixes #

**Any other useful info**:


davestephens (approved):

davestephens:
Good spot, thank you!

PR #564, opened 2022-08-15
FIX: netbootxyz - fix traefik routing and default web page

bcurran3:
* fixed traefik routing 
* pointed traefik to user interface instead of files

Fixes #

**Any other useful info**:


davestephens (approved):

davestephens:
Good spot, thanks :+1:

Issue #526, opened 2022-05-19
How is everyone gettting Nextcloud to upload to .../media/photos?


nicpayne713:
I'm rebuilding my server and cleaning my deployment up... I used to have Nextcloud in Docker volume mount my NAS via SMB (trippy I know) but that way `www-data` could own the mountpoint and I had no permissions issues... This feels super hacky.

Right now I'm trying to setup auto-sync from my phone through nextcloud to the photos directory setup by Ansible-NAS.... I've changed ownership of the target folder inside `photos` to `www-data` but still can't get it working.

Does everyone just keep all their data in the Nextcloud data folder or is there something I'm missing here?

nicpayne713:
I think that description is confusing since my setup isn't known to everyone...

**Simpified Question**

How do I upload via nextcloud to any/all of the samba shares configured by Ansible-NAS?

nicpayne713:
As another update I've realized that I have been trying to use Nextcloud in a manner inconsistent with its design - what I wanted was syncthing or seafile, something to just sync files but not be a more UIX driven solution. I've decided to just use Nextcloud as intended, uploading files to the directories managed by nextcloud

PR #565, opened 2022-08-15
FIX: FreshRSS - add REQUIRED db to task

bcurran3:

- add postgres container for data storage and link to it

REF: Requirements at https://github.com/FreshRSS/FreshRSS/blob/edge/README.md 

Fixes #

**Any other useful info**:

### DIFF
--- a/roles/freshrss/tasks/main.yml
+++ b/roles/freshrss/tasks/main.yml
@@ -5,7 +5,7 @@
     state: directory
     # mode: 0755
   with_items:
-      - "{{ freshrss_data_directory }}/pgsql_data"
+    - "{{ freshrss_data_directory }}/pgsql_data"
     - "{{ freshrss_data_directory }}/data"
     - "{{ freshrss_extensions_directory }}/extensions"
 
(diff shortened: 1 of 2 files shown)


roles/freshrss/tasks/main.yml
```diff
@@ -5,14 +5,34 @@
...
+- name: FreshRSS PostgreSQL Docker Container
+  docker_container:
+    name: freshrss-postgres
+    image: postgres:12-alpine
+    pull: true
+    volumes:
+      - "{{ freshrss_data_directory }}/pgsql_data:/var/lib/postgresql/data"
+    hostname: "freshrss-postgres"
+    ports:
+      - "{{ freshrss_db_port }}:5432"
```
davestephens, line 21: This isn't required in addition to the link in the freshrss definition

davestephens (changes requested):
Just a minor tweak required then should be good to go

roles/freshrss/tasks/main.yml
bcurran3: Sorry but I'm not sure what you are talking about - line #526

bcurran3: If you were talking about the hostname declaration I removed it. Something I have in my files and missed removing for your house style. 
If not - please expound as I'm lost.

PR #568, opened 2022-08-15
FEATURE ADD: Nextcloud - change images to variables

bcurran3:
- create image variables so that users can define/overwrite the defaults

Allow users to choose what images (versions basically) to use for NextCloud. Current default is 14 which is 10 major versions behind the current version. For compatibility sake to OG users the default can remain at v14 and they can then override this when they choose to upgrade. It's recommended that NEW ansible-nas users override the defaults and use the current versions. (Maybe this should be a comment with the variables...hmmmm)

Fixes #

**Any other useful info**:

I know this has been brought up in the past but for whatever reason it hasn't been done. (Maybe because there were other changes that were incorporated at the same time that didn't pass the mustard?) So I figured let's make this simple and get it done.

### DIFF
--- a/roles/nextcloud/defaults/main.yml
+++ b/roles/nextcloud/defaults/main.yml
@@ -5,6 +5,10 @@ nextcloud_available_externally: false
 # directories
 nextcloud_data_directory: "{{ docker_home }}/nextcloud"
 
+# images
+nextcloud_image: "nextcloud:14"
+nextcloud_mysql_image: "mysql:5.7"
+
 # network
 nextcloud_port: "8080"
 nextcloud_hostname: "nextcloud"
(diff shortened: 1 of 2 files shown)


davestephens:
I think this would be good opportunity to flip the image to `latest` too?

bcurran3:
Problem is that I run NC 24 and SQL 8.  So someone else needs to actually test the DB migration that is going to be required for a full frontal upgrade.

I've used this a couple times in the past for upgrades, but not sure if it will work from 14 -> 24:
`docker exec -u 33 nextcloud /var/www/html/occ maintenance:mode --off`

When using the latest tag for NC, newer versions of NC will auto-update. The 10 version jump is problem.

Someone needs to be a Guinea pig and test. Sorry, but I don't have the time to go through it again.

davestephens (approved):

PR #204, opened 2020-01-07
Fix duplicati permissions

animeai:
Fix #171

Issue #171, opened 2019-12-22
Incorrect Duplicati Permissions


h1d3m3:
**Describe the bug**
After a backup using duplicati, you can't restore to the same location because it's mounted ro and missing user ids.

**Environment**
master from Dec 22 : cf86008b101daccfabbe79532145c84f0ecfb3f8

**Expected behavior**
The restore should succeed.

**Actual behavior**
Duplicati fails to restore files with an error that turns out to be a failure to write restored files to their original location.

**Steps to reproduce**
- Backup one of the samba share directories from /source/shares
- Restore those files.

**Additional context**
Original section of duplicati.yml task:

```
    volumes:
      - "{{ duplicati_data_directory }}:/config:rw"
      - "{{ samba_shares_root }}:/source/shares:ro"
      - "{{ docker_home }}:/source/docker:ro"
      - "/etc/timezone:/etc/timezone:ro"
```
I fixed it by adding user/group ids and setting shares and docker to rw:
```
env:
      TZ: "{{ ansible_nas_timezone }}"
      PUID: "{{ nzbget_user_id }}"
      PGID: "{{ nzbget_group_id }}"
    volumes:
      - "{{ duplicati_data_directory }}:/config:rw"
      - "{{ samba_shares_root }}:/source/shares:rw"
      - "{{ docker_home }}:/source/docker:rw"
      - "/etc/timezone:/etc/timezone:ro"
```

animeai:
Fixed with #178

bcurran3:
These changes are simple. I'm willing to whip it out today or tomorrow in like 5 minutes. But,

QUESTION:
```
      - "{{ samba_shares_root }}:/source/shares:rw"
      - "{{ docker_home }}:/source/docker:rw"
```
I see the need to write to the shares so you can put backups someplace such as "dump." I'm just not sure why docker_home should be writable. Let me know so we can move forward.

PR #570, opened 2022-08-15
FIX: duplicati - allow restore of docker persistent data

bcurran3:
- adds ability to restore to docker_home when duplicati_data_permissions is overridden to "rw"

NOTE: I'm not sure why this wasn't addressed in https://github.com/davestephens/ansible-nas/pull/204. I addressed it with a question in https://github.com/davestephens/ansible-nas/issues/171 but never received a response. Hopefully this is a desirable enhancement allowing direct restoring of Docker containers' persistent data without having to restore to a temp area and then move it.

Fixes #

**Any other useful info**:


davestephens (approved):

davestephens:
Not sure why! Thanks for the PR 👍

PR #572, opened 2022-08-15
ADD: DailyTxt - an encrypted Diary Web-App to write down your stories of the day and to find them again easily

bcurran3:

Adds DailyTxt - an encrypted Diary Web-App to write down your stories of the day and to find them again easily

Fixes #

**Any other useful info**:


bcurran3:
I make my kids keep a daily journal in this as it's very basic and simple to use.

Issue #560, opened 2022-08-02
NextCloud non core apt packages


flammableliquids:
Every time I run the playbook and nextcloud is processed, a problem crops up:
I need to re-install the following packages: smbclient libsmbclient libmagickcore-6.q16-6-extra, and then change the config file for magik to allow processing of PDF's. 
One solution I was able to find was to use something like a [docker file](https://stackoverflow.com/questions/47903079/how-to-install-packages-from-docker-compose), but the whole point of ansible is to automate things like this, but I can't seem to figure out how.
This is very frustrating.

**Environment**
- Ansible-NAS revision (`3352773`):
- operating system Ubuntu 20.04.4 LTS
- Docker version 20.10.17, build 100c701
- Running the playbook on the box itself. 
- Vagrant not installed
running on ZFS

davestephens:
What version of Ubuntu are you using?

flammableliquids:
Updated ticket with OS Version (Copy and paste mistake)

flammableliquids:
NextCloud does not install these packages by default, it considers them an optional security risk and recommends you install them yourself.

I guess this is less of a bug, and more of a help / feature request...

davestephens:
You're right, but it's a bug with the nextcloud docker image, not ansible-nas. Installing those packages onto your Ansible-nas will have no effect on what goes on in the nextcloud docker image. 

I'm going to close and recommend re-raising it on the nextcloud repo. It looks like a similar issue [here](https://github.com/nextcloud/docker/issues/633) has been closed with no fix, I have no idea why!

Perhaps there's an alternative nextcloud image we could use, I'd happily accept a PR if a better one can be found. I don't use it myself though so will need to rely on others for testing.

flammableliquids:
The nextcloud core image is "paired down" to not include anything that is of even middling security risk.

davestephens:
It looks like nextcloud provide an all in one image ("aio") - do you want to give that a go and raise a PR if it looks suitable? Alternatively there's a Linuxserver flavour too, their images are usually pretty good.

bcurran3:
> NextCloud does not install these packages by default, it considers them an optional security risk and recommends you install them yourself.
> 
> I guess this is less of a bug, and more of a help / feature request...

This could be done by adding a variable to run a command in the container. The nextcloud setup task could then run the command (i.e. apt install smbclient). The variable could default to "" so that nothing actually runs.

I'll look into doing this in the future.

[...]

PR #576, opened 2022-08-16
FEATURE ADD: Nextcloud - added ability to run custom commands in NC container for configuration post install

bcurran3:

Adds ability to run custom commands in NC container for configuration post install. 

Fixes # Addresses request https://github.com/davestephens/ansible-nas/issues/560

**Any other useful info**:

Breaks AN's CI / Lint rules about shell commands and will have be manually overridden.


davestephens:
After some thought, I'm not going to merge this (or a fixed version of it) as it's not fixing the right problem, and won't work reliably.

The root issue is caused by the Nextcloud containers missing dependencies that Nextcloud requires. Whilst this would hack the problem for the first time Ansible-NAS starts the container, once Watchtower updates the image the problem will return, Nextcloud will be broken again, and I'll end up with more GH issues telling me Nextcloud doesn't work properly.

The right fix is to test and flip to a working Nextcloud container (the Linuxserver one perhaps?), or to petition the Nextcloud container maintainer to add the stuff Nextcloud needs to run without issue.

PR #577, opened 2022-08-16
ADD: WireGuard - an extremely simple yet fast and modern VPN that utilizes state-of-the-art cryptography.

bcurran3:

Adds WireGuard - an extremely simple yet fast and modern VPN that utilizes state-of-the-art cryptography.

Fixes #

**Any other useful info**:

I've been using this container over a year (or two?). Recently I read about another image that has a web configuration front end built into it that I will eventually check out and possibly relace this image with.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ If you have a spare domain name you can configure applications to be accessible 
 * [Virtual Desktop](https://github.com/RattyDAVE/docker-ubuntu-xrdp-mate-custom) - A virtual desktop running on your NAS.
 * [Wallabag](https://wallabag.org/) - Save and classify articles. Read them later.
 * [Watchtower](https://github.com/v2tec/watchtower) - Monitor your Docker containers and update them if a new version is available
+* [WireGuard](https://www.wireguard.com/) - an extremely simple yet fast and modern VPN that utilizes state-of-the-art cryptography.
 * [YouTubeDL-Material](https://github.com/Tzahi12345/YoutubeDL-Material) - Self-hosted YouTube downloader built on Material Design
 * [ZNC](https://wiki.znc.in/ZNC) - IRC bouncer to stay connected to favourite IRC networks and channels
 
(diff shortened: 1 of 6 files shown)


bcurran3:
NOTE: I'm considering changing the image to weejewel/wg-easy but there is another container that just provides a web UI w/o that appears to be generic to work with any WG container. So I'm going to do some testing and see what I think will end up working best (easiest) for most users

davestephens:
Hold / add ? What am I doing with this? Is this ready for review? If it's not ready, please close the PR until it's ready to go. :+1:

bcurran3:
It is ready to go as is, but I **DO** want to evaluate a different image and request it be on hold until I update.

davestephens:
Ok, so it's not ready then.

bcurran3:
After about a year and a half of using the linuxserver image with no problems other than some initial configuration hassles, I've tested, evaluated, and decided that the weejewel/wg-easy image is going to be much easier for the average person to setup and administer. So I changed the image and everything related to it.

With the possible exception of you personally having an issue with the placement of the TZ container variable...

image

georgejung:
Hi I tried to use this to get wireguard up and running but ran into some problems. at first the playbook wasn't working, i think due to ansible dns variable which I dont have defined anywhere. Iam using 1.1.1.1 instead. Then once the container created, I was getting wg0 errors in the docker log

I added to the tasks to incorporate below and that gets me to the web interface with seemingly no errors in the docker log. But I can't seem to get a connection actually working.

cap_add:
      - NET_ADMIN
      - SYS_MODULE
    sysctls:
      - net.ipv4.ip_forward=1
      - net.ipv4.conf.all.src_valid_mark=1
      
I set default address to 10.8.0.1 and then made my clients 10.8.0.2 and 10.8.0.3 and the two clients cant ping each other. I can access the webui from outside but is that the only point of putting the behind traefik? I added an entrypoint in my traefik.toml for udp on port 50821, maybe this was unnecessary? Any advice on how this sohuld work would be great. Ideally i would be able to connect to all of my apps on the ansible-nas but I am struggling.

Thanks

PR #580, opened 2022-08-16
ADD: Readarr - an ebook collection manager for Usenet and BitTorrent users

bcurran3:

Adds Readarr - an ebook collection manager for Usenet and BitTorrent users

...because we can't be missing any of the arr programs now can we?!

Fixes #

**Any other useful info**:

### DIFF
--- a/roles/readarr/defaults/main.yml
+++ b/roles/readarr/defaults/main.yml
@@ -6,6 +6,7 @@ readarr_available_externally: false
 # directories
 readarr_data_directory: "{{ docker_home }}/readarr"
 readarr_download_directory: "{{ downloads_root }}"
+readarr_books_directory: "{{ books_root }}"
 
 # uid / gid
 readarr_user_id: "998"
--- a/roles/readarr/defaults/main.yml
+++ b/roles/readarr/defaults/main.yml
@@ -9,8 +9,9 @@ readarr_download_directory: "{{ downloads_root }}"
 readarr_books_directory: "{{ books_root }}"
 
 # uid / gid
-readarr_user_id: "998"
-readarr_group_id: "1001"
+readarr_user_id: "0"
+readarr_group_id: "0"
+
 # network
 readarr_hostname: "readarr"
 readarr_port: "8787"
--- a/roles/readarr/defaults/main.yml
+++ b/roles/readarr/defaults/main.yml
@@ -1,0 +1,18 @@
+---
+# enable or disable the application
+readarr_enabled: false
+readarr_available_externally: false
+
+# directories
+readarr_data_directory: "{{ docker_home }}/readarr"
+readarr_download_directory: "{{ downloads_root }}"
+
+# uid / gid
+readarr_user_id: "998"
+readarr_group_id: "1001"
+# network
+readarr_hostname: "readarr"
+readarr_port: "8787"
+
+# specs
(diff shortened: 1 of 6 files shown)


roles/readarr/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+    path: "{{ item }}"
+    state: directory
+  with_items:
+    - "{{ readarr_data_directory }}"
+    - "{{ readarr_data_directory }}/config"
+
+- name: Create Readarr Docker Container
+  docker_container:
+    name: readarr
+    image: linuxserver/readarr:nightly
```
davestephens, line 13: Why nightly?

roles/readarr/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+    pull: true
+    volumes:
+      - "{{ readarr_data_directory }}/config:/config"
+      - "{{ books_root }}:/books"
```
davestephens, line 17: Please abstract this the same as `readarr_downloads_root`

roles/readarr/defaults/main.yml
```diff
@@ -0,0 +1,18 @@
...
+# enable or disable the application
+readarr_enabled: false
+readarr_available_externally: false
+
+# directories
+readarr_data_directory: "{{ docker_home }}/readarr"
+readarr_download_directory: "{{ downloads_root }}"
+
+# uid / gid
+readarr_user_id: "998"
```
davestephens, line 11: Where did this uid / gid come from?

davestephens (changes requested):
Few questions and changes required on this PR.

roles/readarr/tasks/main.yml
bcurran3: Only choices are developer and nightly
REF: https://hub.docker.com/r/linuxserver/readarr

roles/readarr/defaults/main.yml
bcurran3: Good catch. 998 is my ansible-nas user. I'll have to go test and figure out why I choose that long ago. Probably should be 1000 if 0 doesn't work.

roles/readarr/tasks/main.yml
bcurran3: done

roles/readarr/defaults/main.yml
bcurran3: updated to use root like all the other *arr linuxserver app defaults

roles/readarr/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+    - "{{ readarr_data_directory }}/config"
+
+- name: Create Readarr Docker Container
+  docker_container:
+    name: readarr
+    image: linuxserver/readarr:nightly
+    pull: true
+    volumes:
+      - "{{ readarr_data_directory }}/config:/config:rw"
+      - "{{ books_root }}:/books:rw"
```
davestephens, line 17: You need to use the new variable you created - I wouldn't expect to see `books_root` here.

davestephens (changes requested):

roles/readarr/tasks/main.yml
bcurran3: done

PR #581, opened 2022-08-16
ADD: DDNS Updater - Container to update DNS records periodically with WebUI for many DNS providers

bcurran3:

Adds DDNS Updater - a container to update DNS records periodically with WebUI for many DNS providers

There was an attempted add of this a year and a half ago that stalled https://github.com/davestephens/ansible-nas/pull/423. 

Fixes #

**Any other useful info**:

I've been using this for about two years and love it. The dev is very personable and is on top of the program with fixes and feature adds.


docs/applications/ddns-updater.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Usage
+
+Set `ddns_updater_enabled: true` in your `inventories/<your_inventory>/nas.yml` file.
+
+The DDNS Updater web interface can be found at <http://ansible_nas_host_or_ip:8001>.
+
+## Specific Configuration
+
+DDNS Updater requires editing its <docker_home>/ddns-updater/data/config.json file. Instructions can be found [here](https://github.com/qdm12/ddns-updater#Configuration).
```
davestephens, line 15: Users shouldn't have to manually meddle with stuff in `<docker_home>/ddns-updater/data/config.json` to get set up. Please add as a yaml variable and template in at deploy.

davestephens (changes requested):

docs/applications/ddns-updater.md
bcurran3: There are 30+ DDNS providers that DDNS Updater supports. Editing the config.json file in the persistent data directory for each provider is very well documented on the dev's site. 
I was going to say that this is impossible to automate, but nothing's impossible. It just would require creating anywhere up to 31 template files since the fields in the json file would have some standard definitions overlap but also contain provider specific fields. On top of that the task itself would need to filter through many conditions based on the variables enabled to pick and generate from the template files. Again, not impossible but a lot of work. IMHO if someone can login to their any of their apps and do basic setup they edit one file in their Docker persistent data dir. Consider how complex and time consuming it is to setup Nextcloud vs editing one text file for this.

If you're hard fast on this, I'd recommend a compromise of getting the app pulled in as is so people can make use of it and then make it an issue to improve upon it.

bcurran3: (When editing the scrutiny.md file, a lightbulb magically appeared above my head...)

To put it in perspective and drive it home: Count the number of times the AN app documentation says 

> edit your `inventories/<your_inventory>/nas.yml` file

image

davestephens: It doesn't need to be that complex. Yaml is JSON, JSON is yaml. 

You could add it a variable, then use Ansible filters to transform into the correct format at template time. A quick Google will reveal lots of examples. There are other examples already in Ansible-NAS. 

A key tenet of Ansible-NAS is that it *automates* setup. If you have to run the playbook then immediately log in to fix something the playbook didn't do, then that tenet is broken (and why I hate nextcloud so much!)

bcurran3: I get it. I know how to do it, I just didn't want to have to do it 31 times!

I'll make the 31 template files...

davestephens:
I think you're misunderstanding. You don't need to make 31 templates, you
don't need to make any.

Create a variable in the role defaults "ddns_config" or something like
that. In that variable is the contents of the config file. Use an ansible
filter to serialise that to json at time of deploy. Done.

Do some googling and you'll see what I mean! I'm pretty sure there's an
example of this already in Ansible-NAS, I just can't find it on mobile.

bcurran3:
Ahhhhhhhhhhh. OK. Got it. Thanks. I understand what you mean now.... basically the same as I learned today about creating a variable with a list... I basically need to add 31 variables to represent each of the 31 possible json files, create another related 31 x 5-7 variables to then be used to generate the actual json file. So almost 200 variables to add. Does that seem right? (OK, slightly less as the variable for e-mail and domain name could be used directly...)

davestephens:
No. Don't concern yourself with what's in the file.

Just add *one* variable (like ddns_config), and then ensure that there's a
task to output that as JSON.

The answer to this question is what I have in mind:
https://stackoverflow.com/questions/63416703/how-to-write-a-json-file-using-ansible

davestephens:
Dupe of #423 , closing.

PR #582, opened 2022-08-16
ADD: scrutiny - Hard Drive S.M.A.R.T Monitoring, Historical Trends & Real World Failure Thresholds

bcurran3:

**NOTE: This is a work in progress and help is requested to overcome a configuration obstacle.**

Adds scrutiny - Hard Drive S.M.A.R.T Monitoring, Historical Trends & Real World Failure Thresholds

**Which issue (if any) this PR fixes**: https://github.com/davestephens/ansible-nas/issues/2 from the very early days of AN!

Fixes #

**Any other useful info**:

This scrutiny addition has a problem to overcome that I'm not technically proficient enough at the moment to solve. I'm requesting help to get this finished. The problem is the list of devices (disks) will be different for different users. Thus the devices can't statically be assigned, with the possible exception of sda, in the task file. Asking users to edit the task file is out of the question as it would just be overwritten on their next git pull. So the disk device list needs to be created dynamically at runtime. Ansible facts can supply the info. A little Googlefu and I can get the info, but I (repeat) am not technically proficient enough at Ansible to process the info arrays and use it. So hopefully someone good at that stuff can help finish this.

This gets and displays the needed info, but that's as far as I got.
```
################################ TESTING ########################################
- name: Output SATA disks
  debug:
    var: hostvars[inventory_hostname].ansible_devices.keys() | map('regex_search', 'sd.*') | select('string') | list

- name: Output NVME disks
  debug:
    var: hostvars[inventory_hostname].ansible_devices.keys() | map('regex_search', 'nvme.*') | select('string') | list
################################ TESTING ########################################
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ If you have a spare domain name you can configure applications to be accessible 
 * [Route53 DDNS](https://crazymax.dev/ddns-route53/) - Automatically update AWS Route53 with your IP address
 * [RSS-Bridge](https://rss-bridge.github.io/rss-bridge/) - The RSS feed for websites missing it
 * [Sabnzbd](https://sabnzbd.org/) - A powerful usenet downloader that FreeNAS provides
+* [scrutiny](https://github.com/AnalogJ/scrutiny) - Hard Drive S.M.A.R.T Monitoring, Historical Trends & Real World Failure Thresholds
 * [Sickchill](https://sickchill.github.io/) - for managing TV episodes
 * [Sonarr](https://sonarr.tv/) - for downloading and managing TV episodes
 * [Speedtest-Tracker](https://github.com/henrywhitaker3/Speedtest-Tracker) - Continuously track your internet speed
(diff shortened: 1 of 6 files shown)


bcurran3:
NOTE: The container documentation says to open a port for Influxdb. I think it's an error as Influxdb is installed in the container along with scrutiny - they they can talk directly without leaving the container's network. With the port open you can access the Influxdb's web UI, but no credentials can be found. I've closed the port, and thus direct access to Influxdb, and scrutiny continues to work flawlessly. I've purposely left scrutiny_influxdb_port in the defaults file just in case I'm in error or something changes in the future but am not using it in the task.

bcurran3:
Oh cool! I think I just discovered the way to solve this! Yay! Excited!

I didn't realize that an Ansible variable can contain a list. Just discovered this is the ansible-nas-general defaults. Super happy. I think this will solve the problem and I'll work on it later today or tomorrow.

REF:
```
# Extra packages to install
ansible_nas_extra_packages:
  - smartmontools
  - htop
  - zfsutils-linux
  - bonnie++
  - unzip
  - lm-sensors
  - ctop
```

bcurran3:
Problem overcome. Updated. Tested. Ready for review.

bcurran3:
Please re-run checks. GH has been having problems.

> Got 1 exit code while running: ansible-galaxy role install -vr requirements.yml --roles-path /home/runner/.cache/ansible-compat/ffe364/roles

PR #586, opened 2022-08-18
FIX: Speedtest-Tracker - correct container name

bcurran3:

Fixes the Speedtest-Tracker container name from speedtest_tracker to speedtest-tracker

Yes, it's minor, but it's proper.

Fixes #

**Any other useful info**:

This will cause a bind failure if the container is already running as speedtest_tracker. I am willing to add code to remove (state : absent) the original container if you so desire.

### DIFF
--- a/roles/speedtest-tracker/tasks/main.yml
+++ b/roles/speedtest-tracker/tasks/main.yml
@@ -8,7 +8,7 @@
 
 - name: Speedtest-Tracker Docker Container
   docker_container:
-    name: speedtest_tracker
+    name: speedtest-tracker
     image: henrywhitaker3/speedtest-tracker:latest
     pull: true
     volumes:
--- a/roles/speedtest-tracker/tasks/main.yml
+++ b/roles/speedtest-tracker/tasks/main.yml
@@ -10,7 +10,7 @@
   docker_container:
     name: speedtest_tracker
     state: absent
-  when: speedtest_tracker_enabled = true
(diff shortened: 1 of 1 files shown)


roles/speedtest-tracker/tasks/main.yml
```diff
@@ -8,7 +8,7 @@
 
 - name: Speedtest-Tracker Docker Container
   docker_container:
-    name: speedtest_tracker
+    name: speedtest-tracker
```
davestephens, line 11: If someone is already running it, this change will cause the new container to fail to start. 

A step is required prior to this to stop/remove the old one - which is arguably more mess than just leaving this how it is.

davestephens (changes requested):

roles/speedtest-tracker/tasks/main.yml
```diff
@@ -6,9 +6,15 @@
   with_items:
     - "{{ speedtest_data_directory }}/config"
 
-- name: Speedtest-Tracker Docker Container
+- name: Remove old Speedtest-Tracker Docker Container
   docker_container:
     name: speedtest_tracker
+    state: absent
+  when: speedtest_tracker_enabled == true
```
davestephens, line 13: Aside from breaking the build, this line isn't required as it's covered by the `speedtest_tracker_enabled` applied at the role level.

davestephens (changes requested):

davestephens (approved):

PR #589, opened 2022-08-20
FIX: Traefik - add local TZ environmental variable

bcurran3:

Adds local TZ variable to Traefik container.

Looking at logs is very tedious for those of us not living in UTC!

Fixes #

**Any other useful info**:

### DIFF
--- a/roles/traefik/defaults/main.yml
+++ b/roles/traefik/defaults/main.yml
@@ -20,6 +20,7 @@ traefik_log_level: "INFO"
 # find the relevant name and environment variables for your DNS provider at https://go-acme.github.io/lego/dns/
 traefik_dns_provider: cloudflare
 traefik_environment_variables:
+  TZ: "{{ ansible_nas_timezone }}"
   CF_DNS_API_TOKEN: "abcdabcd123412341234"
 
 # Ansible-NAS requests a wildcard certificate for your domain, so there should be no reason to have to use the staging


roles/traefik/defaults/main.yml
```diff
@@ -20,6 +20,7 @@ traefik_log_level: "INFO"
 # find the relevant name and environment variables for your DNS provider at https://go-acme.github.io/lego/dns/
 traefik_dns_provider: cloudflare
 traefik_environment_variables:
+  TZ: "{{ ansible_nas_timezone }}"
```
davestephens, line 23: This should be added directly to the `docker_container` definition as with other containers. It'll work how it is, but is not in the right place.

davestephens (changes requested):

roles/traefik/defaults/main.yml
bcurran3: Trying to do this:

```
      - TZ: "{{ ansible_nas_timezone }}"
      - "{{ traefik_environment_variables }}"
```
Results in this:

> TASK [traefik : Traefik Docker Container] ******************************************************************************
> fatal: [ansible-nas]: FAILED! => {"changed": false, "msg": "argument env is of type <class 'list'> and we were unable to convert to dict: <class 'list'> cannot be converted to a dict"}

PR #590, opened 2022-08-20
ENHANCMENT: Traefik - add ability to use rules.toml file

bcurran3:

Adds ability to use a rules.toml file to redirect to websites outside of Ansilbe-NAS

Fixes #

**Any other useful info**:

### DIFF
--- a/roles/traefik/tasks/main.yml
+++ b/roles/traefik/tasks/main.yml
@@ -5,12 +5,13 @@
     state: directory
   with_items:
     - "{{ traefik_data_directory }}"
+    - "{{ traefik_data_directory }}/data"
     - "{{ traefik_data_directory }}/letsencrypt"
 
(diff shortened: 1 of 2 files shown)


davestephens:
This doesn't provide any way to customise the file referenced in the config. Is this PR complete?

davestephens (changes requested):
This doesn't provide any way to customise the file referenced in the config. Is this PR complete?

bcurran3:
Subjective. I would say yes. I considered a template but the possibilities to be in that template are endless; it's too open. Plus if you edit roles.toml directly in a template setup, it'll be overwritten. With that said  a list variable to insert into the template might work.

davestephens:
The config reference will load the file from within the container. What puts it there?

davestephens:
Closing as not something I want to support

PR #591, opened 2022-08-20
FIX: Traefik - traefik_port_http variable defined but never used

bcurran3:

In the switch to a role, the variable traefik_port_http was created in the defaults but not used. This implements it's intended purpose.

Fixes #

**Any other useful info**:

### DIFF
--- a/roles/traefik/templates/traefik.toml
+++ b/roles/traefik/templates/traefik.toml
@@ -1,6 +1,6 @@
 [entryPoints]
   [entryPoints.web]
-    address = ":traefik_port_http"
+    address = ":{{ traefik_port_http }}"
 
   [entryPoints.web.http.redirections.entryPoint]
     to = "websecure"
--- a/roles/traefik/templates/traefik.toml
+++ b/roles/traefik/templates/traefik.toml
@@ -1,6 +1,6 @@
 [entryPoints]
   [entryPoints.web]
-    address = ":80"
+    address = ":traefik_port_http"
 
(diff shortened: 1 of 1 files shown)


roles/traefik/templates/traefik.toml
```diff
@@ -1,6 +1,6 @@
 [entryPoints]
   [entryPoints.web]
-    address = ":80"
+    address = ":traefik_port_http"
```
davestephens, line 3: This isn't an ansible variable, hence won't be templated

davestephens (changes requested):

davestephens:
PRs are great, but please test the code first.

roles/traefik/templates/traefik.toml
bcurran3: You should take another look at:
https://github.com/davestephens/ansible-nas/blob/master/roles/traefik/defaults/main.yml

bcurran3:
> PRs are great, but please test the code first.

You're 100% right. 

Unfortunately as far as Traefik goes I've been 100% effed for the last two weeks. 

I've been fervently going over all the Traefik code and docs to figure out what I've got wrong and can't figure it out. Frustration levels are high enough that I would even consider asking for help. :-) BUT don't want to see "RTFM" told to me. Do you personally have a few minutes to look over my config (I'd send in e-mail) and possibly see what I did wrong?

davestephens (approved):

Issue #432, opened 2021-02-25
Hardening SSL


georgejung:
**Is your feature request related to a problem? Please describe.**
Want to have something as secure as is possible, being usable. SSL Check by default gives me B rating (https://www.ssllabs.com/ssltest/) due to TLS 1, and weak ciphers.

**Describe the solution you'd like**
Integrate dynamic config file into the role. The minimum TLS version and accepted ciphers can be set in a dynamic_conf.toml file. If we put that in our templates folder, reference it in our traefik.toml, and update the tasks/main.yml then we can harden the connection significantly, and I have not lost any functionality.

I was able to set the minimum TLS version but the ciphers hasn't seemed to work for me yet in traefik 2.4. I had both minimium tls and ciphers working in my traefik.toml (v1.7).

**Describe alternatives you've considered**
None

**Additional context**
https://github.com/georgejung/ansible-nas/tree/master/roles/traefik

I forked and made the changes for this specific issue in the above repo. I was hesitant to do a pull request since I didn't do all of the testing required, but that is my working code, for whatever its worth.

Thanks again for everyone's work on this project.

HitLuca:
@georgejung You made the same mistake as me, as not mounting the new dynamic configuration in the traefik docker container!

PR #595, opened 2022-08-21
Forced traefik to only allow tls1.2+ requests

HitLuca:
Updated the traefik dynamic config to force clients to use tls1.2 or newer

### DIFF
--- a/roles/traefik/tasks/main.yml
+++ b/roles/traefik/tasks/main.yml
@@ -7,11 +7,17 @@
     - "{{ traefik_data_directory }}"
     - "{{ traefik_data_directory }}/letsencrypt"
 
-- name: Template Traefik config.toml
+- name: Template Traefik traefik.toml
   template:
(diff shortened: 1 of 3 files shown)


HitLuca:
Partially fixes #432, doesn't take into account cyphers

roles/traefik/tasks/main.yml
```diff
@@ -7,12 +7,18 @@
...
   template:
     src: traefik.toml
     dest: "{{ traefik_data_directory }}/traefik.toml"
   register: template_config
 
+- name: Template Traefik dynamic_config.toml
+  template:
+    src: dynamic_config.toml
+    dest: "{{ traefik_data_directory }}/dynamic_config.toml"
+  register: template_dynamic_config
```
kamilkrz, line 20: As in this manner dynamic config is static file should it be `file:` instead of `template:`

PR #596, opened 2022-08-21
Added code-server app

### LINKED ISSUE #534: Add VSCode Server image
Used for remote development, I think this app would be a nice addition to this project.

The site for the docker image is [this](https://coder.com/docs/code-server/latest), and the setup guide for docker is [here](https://coder.com/docs/code-server/latest/install#docker).

The setup is pretty straightforward, In my case I created the container using Portainer, and mounted my projects folder from a zpool dataset. 

Here is my setup if it can help design the feature:

```yaml
version: '3.7' 
services: 
    vscode_server: 
        image: codercom/code-server
        container_name: vscode_server 
        restart: unless-stopped 
        ports:
            - 9050:8080 #<- the vscode image uses port 8080, so we need to change it since nextcloud uses it
        volumes: 
            - ${CONFIG_FOLDER}:/home/coder/.config
            - ${PROJECTS_FOLDER}:/home/coder/project
        environment: 
            - TZ=${TZ}
            - DOCKER_USER=$USER            
```

HitLuca:
Added code-server image, which is VSCode server without the Microsoft branding. Also it is open source, I prefer to add this instead of VSCode honestly.

Still a work in progress, we are missing docs and app ports sections for code-server, the rest is there and has been tested.

It also adds a `code_root` variable, which is where the projects would be stored. I chose it to be `samba_shares_root/code`

Fixes #534

**Any other useful info**:
I noticed that since code-server has a PUID/PGID of 0, all files created are owned by root. Changing the ids to, for example, the ones of `ansible_nas_user` would be better in my opinion


HitLuca:
I just remembered one thing: it's better to have the choice of using code-server and vscode-server, since the open source code-server doesn't have all the extensions available in the marketplace.

If it's ok for you I can make another pr for vscode-server tomorrow

davestephens (changes requested):

HitLuca:
Before it was vscode-server, when I refactored the name I forgot to reorder it lol

HitLuca:
> I just remembered one thing: it's better to have the choice of using code-server and vscode-server, since the open source code-server doesn't have all the extensions available in the marketplace.

Well forget what I said, I was thinking of `Codium`, which is VSCode without branding, not `code-server` vs `vscode-server` my bad. There's no `vscode-server` docker image, only `code-server`, which is VSCode containerized. More details [here](https://www.roboleary.net/tools/2022/05/31/vscode-vscodium-which-should-i-use.html)

I'll just add docs for this and the PR is good to go

davestephens (approved):

PR #597, opened 2022-08-21
Removed grafana and telegraf related files, moved users.yml from task to role

HitLuca:
I did some cleanup and removed the leftover files related to grafana and telegraf, as well as moving users.yml from being the last task to being a role as everything else

### DIFF
--- a/roles/traefik/templates/traefik.toml
+++ b/roles/traefik/templates/traefik.toml
@@ -1,6 +1,6 @@
 [entryPoints]
   [entryPoints.web]
-    address = ":80"
+    address = ":{{ traefik_port_http }}"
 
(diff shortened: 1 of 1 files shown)


davestephens (changes requested):
I've actually been working on reviving the stats role this evening - nearly ready to merge :-)

Please can you remove the stats cleanup stuff and it'll leave the ansible-nas-users role in this PR on its own.

HitLuca:
@davestephens made a new PR with only the users change. I would suggest avoiding leaving files in the repo "because I'm going to use them later" though, master/main should all be clean of unused stuff, and deleted files can be brought back from an earlier commit if needed via cherrypick or similar

PR #598, opened 2022-08-21
WIP: ENHANCEMENT - YouTubeDL-Material - add UID/GID env options

bcurran3:

 YouTubeDL-Material - add UID/GID env options

Not sure if the container supported this when I first added this to AN, but it does now. I left the defaults for UID/GID at the dev's defaults but now they can be overwridden.

Fixes #

**Any other useful info**:

I've had no success setting it to run as root:

> WARNING! Could not change directory ownership. If you manage permissions externally this is fine, otherwise you may experience issues when downloading or deleting videos.
> error: failed switching to " 0: 0": unable to find user  0: no matching entries in passwd file
> find: ' 0' is not the name of a known user

It says "this is fine" but the container reboots. I have not tried setting to other UID's or GUD's, but now the option is there to play with. :->

### DIFF
--- a/roles/youtubedlmaterial/defaults/main.yml
+++ b/roles/youtubedlmaterial/defaults/main.yml
@@ -9,6 +9,10 @@ youtubedlmaterial_dl_audio_directory: "{{ downloads_root }}/youtube/audio"
 youtubedlmaterial_dl_video_directory: "{{ downloads_root }}/youtube/video"
 youtubedlmaterial_dl_subscriptions_directory: "{{ downloads_root }}/youtube/subscriptions"
 
+# uid / gid
+youtubedlmaterial_user_id: "1000"
+youtubedlmaterial_group_id: "1000"
+
 # network
 youtubedlmaterial_hostname: "youtubedlmaterial"
 youtubedlmaterial_port_http: "8998"
--- a/roles/youtubedlmaterial/tasks/main.yml
+++ b/roles/youtubedlmaterial/tasks/main.yml
@@ -28,6 +28,8 @@
       - "{{ youtubedlmaterial_port_http }}:17442"
     env:
       ALLOW_CONFIG_MUTATIONS: "true"
+      UID: " {{ youtubedlmaterial_user_id }}"
+      GID: " {{ youtubedlmaterial_group_id }}"
       TZ: "{{ ansible_nas_timezone }}"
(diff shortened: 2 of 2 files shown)


davestephens:
What the dev thinks the uid/gid is not relevant in this context. It needs to be set to either 0 for root, or the uid/gid for the ansible Nas user needs to be calculated and used. Depends on the desired behaviour. 

In Ubuntu uid/gid 1000 is the first user added user, ie likely you, which is definitely not the right one.

roles/youtubedlmaterial/defaults/main.yml
```diff
@@ -9,6 +9,10 @@ youtubedlmaterial_dl_audio_directory: "{{ downloads_root }}/youtube/audio"
 youtubedlmaterial_dl_video_directory: "{{ downloads_root }}/youtube/video"
 youtubedlmaterial_dl_subscriptions_directory: "{{ downloads_root }}/youtube/subscriptions"
 
+# uid / gid
+youtubedlmaterial_user_id: "1000"
```
davestephens, line 13: This doesn't look right to me.

roles/youtubedlmaterial/defaults/main.yml
```diff
@@ -9,6 +9,10 @@ youtubedlmaterial_dl_audio_directory: "{{ downloads_root }}/youtube/audio"
...
+youtubedlmaterial_group_id: "1000"
```
davestephens, line 14: Or this

roles/youtubedlmaterial/tasks/main.yml
```diff
@@ -28,6 +28,8 @@
       - "{{ youtubedlmaterial_port_http }}:17442"
     env:
       ALLOW_CONFIG_MUTATIONS: "true"
+      UID: " {{ youtubedlmaterial_user_id }}"
```
davestephens, line 31: Unintentional whitespace next to the uid/gid opening quotes

davestephens (changes requested):
What the dev thinks the uid/gid is not relevant in this context. It needs to be set to either 0 for root, or the uid/gid for the ansible Nas user needs to be calculated and used. Depends on the desired behaviour.

In Ubuntu uid/gid 1000 is the first user added user, ie likely you, which is definitely not the right one.

PR #600, opened 2022-08-23
Fixed linting errors

HitLuca:
Where do these new linting errors come from? The failing commit doesn't seem to have introduced new upgrades for the `ansible-lint` package, which is the one complaining. We can also avoid this issue by adding `name` to the `skip_list` of `.ansible-lint` I believe

### DIFF
--- a/roles/emby/tasks/main.yml
+++ b/roles/emby/tasks/main.yml
@@ -6,7 +6,7 @@
   with_items:
     - "{{ emby_config_directory }}"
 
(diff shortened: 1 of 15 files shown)


HitLuca:
Oh ok the git workflow doesn't pin the `ansible-lint` version. Watch out for stuff like this

davestephens:
Sorry, I saw this after I fixed it myself!

PR #606, opened 2022-09-09
FIX: Deluge conditional fails parsing

fetherolfjd:

The `when:` clause had an additional`when` which needs removal to avoid failure.

No issue created.

**Any other useful info**:


davestephens (approved):

davestephens:
Good find, thanks!

PR #608, opened 2022-10-05
Add: Calibre eBook Manager

### LINKED ISSUE #297: add Calibre proper
(SORRY I used the wrong template!)

**Is your feature request related to a problem? Please describe.**

No.

**Describe the solution you'd like**

We have Calibre-web, seems we should have Calibre proper to go with it!
(There, I spoke British!)

**Describe alternatives you've considered**

I'm a fan of Ubooquity...

**Additional context**

https://hub.docker.com/r/linuxserver/calibre

If no one else does it, I probably will even though I don't use it.

eniad:

Add Calibre eBook Manager

Fixes #297 

**Any other useful info**:

1. I've been running this locally for about two months.
2. Passing the `security_opts` into the docker container fixed this recent error: https://github.com/linuxserver/docker-calibre/issues/102. Glances and netdata are the only other roles to use security options.
3. Verify I correctly added the molecule tests.
4. The implementation of mounting `- "{{ calibre_data_directory }}/data:/data"`  in `roles/calibre/tasks/main.yml` works, but it should be checked. The may be a better way to expose the Calibre Database Directory to the `calibre` docker container.
5. I would like to see the application_ports page return. I was really helpful to see all the used ports in one list.


eniad:
At first glance, `Error: Dependencies lock file is not found in /home/runner/work/ansible-nas/ansible-nas. Supported file patterns: yarn.lock` doesn't feel like something I did in this PR. I'll be glad to make any changes need to get past it, but I don't know how to solve it.

eniad:
> i think it would be better to let this be `calibre` and the existing `calibre` renamed to `calibre-web`.

I agree. However that would break existing installations. I think avoiding breaking changes is more important than perfect naming.

I actually have a local branch that renames calibre-web just as you suggest. If the group decides that renaming is a better action, I can update that branch.

davestephens:
I agree, and would be happy to accept that as part of this PR 👍

eniad:
Rename done in new commit [1e3c150](https://github.com/davestephens/ansible-nas/pull/608/commits/1e3c150fffb2fbc4d9914c3000bbf65d495ec7e2)

I went with `calibreweb` because of [YAML Syntax](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html). To be consistent with Ansible-NAS, I also omitted the hyphen from the role and container names.

eniad:
A note on transitioning the role name:

I had to manually remove the `calibre` docker container before running the playbook with the new role to clear port allocation. That should be a one time fix, but anyone using calibre-web may run into that error.

jakoblover:
Any updates on this?

eniad:
Merge conflict resolved.

davestephens:
MAssively late to the party but finally got this merged - thanks!

PR #619, opened 2022-12-05
Add pre_task veryfing minimal ansible version

vanklompf:
On gitter chat there is number of people having issue with first installation of ansible-nas (including me). In many cases it is incorrect version of ansible. Currently error message is not very helpful as it is showing rather syntax error and not version mismatch. This change is based on: [this page](https://www.jeffgeerling.com/blog/2018/require-minimum-ansible-version-your-playbook)

**What this PR does / why we need it**: adds pre-check for minimal version of Ansible

### DIFF
--- a/nas.yml
+++ b/nas.yml
@@ -1,6 +1,12 @@
 ---
 - hosts: all
 
+  pre_tasks:
+    - name: Verify Ansible meets ansible-nas version requirements.
+      assert:
+        that: "ansible_version.full is version_compare('2.10', '>=')"
+        msg: "update Ansible to at least 2.10 to install ansible-nas"
+
   roles:
     ###
     ### Requirements


davestephens:
Please can you provide more information on the issue you mention?

vanklompf:
When running playbook on slightly older WSL2 or Ubuntu 20.04 (ansible 2.9.6) it fails in the middle with rather misleading error message:
```
TASK [airsonic : Create Airsonic Directories] **********************************************************************************************************************
fatal: [ansible-nas]: FAILED! => {"msg": "The conditional check 'airsonic_enabled is true' failed. The error was: template error while templating string: no test named 'true'. String: {% if airsonic_enabled is true %} True {% else %} False {% endif %}\n\nThe error appears to be in '/root/ansible-nas/roles/airsonic/tasks/main.yml': line 4, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  block:\n    - name: Create Airsonic Directories\n      ^ here\n"}
```

With my change it fails early with more useful error:
```
TASK [Verify Ansible meets ansible-nas version requirements.] ******************************************************************************************************
fatal: [ansible-nas]: FAILED! => {
    "assertion": "ansible_version.full is version_compare('2.10', '>=')",
    "changed": false,
    "evaluated_to": false,
    "msg": "update Ansible to at least 2.10 to install ansible-nas"
}
```
The same playbook works on Ubuntu 22.04 (ansible 2.10.8)

PR #627, opened 2022-12-28
Fix nomad when nomad_enabled is false

nilshamerlinck:
Ignore errors if unit doesn't exist


davestephens:
Closing as now fixed without ignoring errors.

PR #632, opened 2023-01-05
Add Authelia

anarion80:
Adds Authelia

None
Fixes #
None
**Any other useful info**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ If you have a spare domain name you can configure applications to be accessible 
 ## Available Applications
 
 * [Airsonic](https://airsonic.github.io/) - catalog and stream music
+* [Authelia](https://www.authelia.com/) - The Single Sign-On Multi-Factor portal for web apps
 * [Bazarr](https://github.com/morpheus65535/bazarr) - companion to Radarr and Sonarr for downloading subtitles
 * [Bitwarden_rs](https://github.com/dani-garcia/bitwarden_rs) - Self-Hosting port of password manager
 * [Booksonic](https://booksonic.org/) - The selfhosted audiobook server
--- a/nas.yml
+++ b/nas.yml
@@ -45,6 +45,10 @@
       tags:
         - airsonic
 
+    - role: authelia
+      tags:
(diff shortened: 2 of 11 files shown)


davestephens (changes requested):
Thanks for the contribution! With the minimal description in the PR I'm struggling to piece together how this would work, or what the PR provides.

Few thoughts:
- There are numerous hard-coded assumptions (ie whitelisted network ranges)
- Why is there an additional docker network created?
- Why are some of the containers named "example"? 
- I don't like the "edit the configuration.yml" instruction - it's a big ansible-nas anti-pattern. As soon as people do that, it creates merge issues in the future. Also, the file is a template.

Currently, I'm not sure how we can proceed with this in a way that it's useful for people - perhaps if you can explain the thinking it'd help?

anarion80:
Yeeees, I was going alphebetically, and this difficult container was one of the first on the list ;)

* The hardcoded assumptions are not needed, I removed them.
* Docker network is needed since Authelia will act as a middleware for some of the other containers, so those containers to be secured by authelia need to be able to reach authelia
* Containers are named `example` following the official example: https://github.com/authelia/authelia/blob/master/examples/compose/local/docker-compose.yml. They are not really needed.
* Yes, agreed. One would need to create many many more Ansible variables to truly configure Authelia or template the configuration file. As it is now, is rather a good starting point for advanced uers.

To be able to truly integrate Authelia using traefik labels, the whole playbook would probably need to be reworked, Authelia set up as some high-level condition, etc. I'm not willing to go there now, so feel free to close this.

lpkampen:
Hi, I agree that Authelia should be part of this project, I\ve had it running now since february last year.
Did you test what you committed and no other settings are needed on each app?
I haven't looked too closely on your coding but I added this to all my instances I wantet to reach externally:
```
traefik.http.routers.overseerr.entryPoints: websecure
traefik.http.routers.overseerr.middlewares: auth@file
```
I also set it up with different levels of security so that my router, codeserver and a few other apps needs multiple factors for login and other only username and password.

PR #634, opened 2023-01-08
Add Flame

anarion80:
Adds Flame - a self-hosted start page/dashboard

Fixes #

**Any other useful info**:


davestephens (changes requested):
This is looking good - will require a few tweaks then we should be GTG 👍

PR #635, opened 2023-01-08
Add Grocy

anarion80:
**What this PR does / why we need it**:
Adds Grocy - a web-based self-hosted groceries & household management solution for your home.

**Which issue (if any) this PR fixes**:

Fixes #

**Any other useful info**:

PR #636, opened 2023-01-08
Add Barcode Buddy

anarion80:
Adds Barcode Buddy - Barcode system for Grocy 

Fixes #

**Any other useful info**:
This is related to Grocy and is therefore dependent on it, and on https://github.com/davestephens/ansible-nas/pull/635 being available already.


davestephens:
Can we move this into grocy as a configurable option? Given it's dependent on it, it doesn't make sense to have it as a separately installable role.

anarion80:
Yes, I will add it as an option to Grocy

anarion80:
Updated https://github.com/davestephens/ansible-nas/pull/635 with Barcode Buddy as an option

PR #638, opened 2023-01-09
Fix #621

### LINKED ISSUE #621: Playbook run fails attempting to stop nomad
**Describe the bug**
The playbook run ends with a failure attempting to stop nomad.

```
TASK [nomad : Start Nomad] **************************************************************************************************************************
skipping: [ansible-nas]

TASK [nomad : Stop Nomad] ***************************************************************************************************************************
fatal: [ansible-nas]: FAILED! => {"changed": false, "msg": "Could not find the requested service nomad: host"}

PLAY RECAP ******************************************************************************************************************************************
ansible-nas                : ok=78   changed=5    unreachable=0    failed=1    skipped=122  rescued=0    ignored=0 
```

**Expected behavior**
It should not error.

**Actual behavior**
What actually happens?

**Steps to reproduce**
Don't have nomad enabled and run `nsible-playbook -i inventories/<your-nas>/inventory nas.yml -b -K`

anarion80:
Issue #621 still seems to be valid. Looks like Nomad is using the new approach of enabling applications, meaning it is not behind `when: (nomad_enabled | default(False))`, thus the role is always run, thus the Stop Nomad task is always run. And unlike all the other cases where we are stopping a docker container, here we are stopping a systemd service. We have `nomad_enabled: false` by default, but the service is not there and hence the error. So like @fetherolfjd suggests in the issue, we can check if the service is present in the conditional.

This is my proposed fix, maybe a little expanded to actually remove the service, not just stop (Since enabling the flag to `true` is actually installing and enabling the service, so disabling the flag should clean up everything?)

I tested it only using `-t nomad`, and at least there the error is gone.

Fixes #621

**Any other useful info**:

### DIFF
--- a/roles/nomad/tasks/main.yml
+++ b/roles/nomad/tasks/main.yml
@@ -1,4 +1,7 @@
 ---
+- name: Get the list of services
+  service_facts:
+
 - name: Install and start Nomad
   block:
     - name: Add Hashicorp GPG key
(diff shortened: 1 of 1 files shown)


fetherolfjd:
#627 was opened a few weeks ago too, which just ignores any errors. I do like the idea of handling this more gracefully though.

davestephens:
I attacked this a slightly different way so closing this. Thanks for the idea in your PR!

PR #639, opened 2023-01-09
add note about firewall settings

StevenCHowell:

Update the Plex documentation, outlining the need to and resources for changing the firewall settings.

### DIFF
--- a/website/docs/applications/media-serving/plex.md
+++ b/website/docs/applications/media-serving/plex.md
@@ -16,3 +16,7 @@ The Plex web interface can be found at <http://ansible_nas_host_or_ip:32400/web/
 
 You'll need to log in with a plex account, (or possibly claim your server).
 You'll need to set up your local file store - go to settings > Manage > Libraries, and add your libraries. These are mounted in the root directory by default.
+
+### Firewall Settings
+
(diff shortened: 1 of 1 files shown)


davestephens:
I'm not clear which firewall you're referring to. Do you mean a user's personal router firewall, or a UFW firewall on the Ansible-NAS box? The latter article you link refers to UFW, which isn't enabled on Ubuntu 22.04 by default...hence my confusion.

viktor-c:
> I'm not clear which firewall you're referring to. Do you mean a user's personal router firewall, or a UFW firewall on the Ansible-NAS box? The latter article you link refers to UFW, which isn't enabled on Ubuntu 22.04 by default...hence my confusion.

Just used ansible-nas to install some containers on a freshly installed Ubuntu 22.04, which only got configured to use ssh. The ubuntu-installer gives the option to install docker, which I did. This proved to be a mistake, because the ansible playbook failed to install usable containers. Dokuwiki worked through traefik, but containers supposed to talk to each other could not connect. I thought it could be a problem with ubuntu firewall: ufw. Allowing ports was not enough. More than that, I had a freshly configured hetzner server, on which I ran the playbook. This worked flawlessly, inspite both servers having same ubuntu, kernel, docker, docker compose and ansible versions installed. 

So the problem was during install of my physical server, when I choose  to install docker through it. This installs a snap? version of docker, which does not work well with ansible-nas playbook....
Maybe this helps somebody, cost me 1 day and some gray hair...

davestephens:
OK so the actual issue is that Docker shouldn't be installed before running Ansible-NAS. I've updated the docs accordingly.

PR #650, opened 2023-02-15
Fix docs test

anarion80:

This is an attempt to fix the "Test Website Deployment" action that always fails currently. I'm not sure why it was using `yarn` when the Deploy action uses `npm`, so I switched it to `npm` for consistency. But is also works with `yarn` after adding `cache-dependency-path` and default working directory.

Fixes #

**Any other useful info**:


davestephens:
Thanks, this has been on my list to look at for a while 👍

PR #654, opened 2023-03-06
:sparkles: Add badge with role count

anarion80:

Adds a badge that shows the number of roles/applications. It counts the number of directories in `roles` directory. The count is approximate (we have 3 internal roles), but still it's nice ;)

Fixes #

**Any other useful info**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # Ansible NAS
 
+![Available Applications](https://img.shields.io/github/directory-file-count/davestephens/ansible-nas/roles?label=Available%20Applications&style=flat-square)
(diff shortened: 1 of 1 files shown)


davestephens:
This is a really neat idea! Please move it to the very top of the readme though, ie, the first badge, and rename it "Available Applications", and add the `flat-square` style to match the other badges.

davestephens (changes requested):
Please change per the PR comment :+1:

anarion80:
Updated as per PR comment

davestephens (approved):

PR #658, opened 2023-04-28
:sparkles: Add Admidio

anarion80:

Adds Admidio.

NOTE: This install fine and is working, but there seems to be possibly an issue with Admidio server itself: https://github.com/Admidio/admidio/issues/1404
so this is currently not yet usable.

Fixes #

**Any other useful info**:


anarion80:
This was supposed to go to my fork, sorry.

PR #667, opened 2023-07-28
Add support for photoprism

guipguia:
add support for deploying photoprism

**Any other useful info**:
Can also close this old PR: https://github.com/davestephens/ansible-nas/pull/543


darshitpp:
@davestephens Can we add the support for this? I'd like Photoprism as Piwigo looks comparatively aged.

PR #672, opened 2023-08-31
Update mkdocs.yml year

Stihotvor:
Change the year

Changes the year to 2023

Fixes #

**Any other useful info**:


davestephens:
Not using mkdocs any more...Thanks though!

PR #674, opened 2023-09-17
Update Transmission Role External Port

jaimesBooth:
Update Role Transmission external port to 51413 to match external port defined in Task so that Listening Port can be "open"

 Update Transmission Role External Port: updates Role Transmission external port to 51413 to match external port defined in Task so that Listening Port can be "open"

Fixes #

**Any other useful info**:
Was unable to detect open listening port within Transmission without matching ports and forwarding in router.

### DIFF
--- a/roles/transmission/defaults/main.yml
+++ b/roles/transmission/defaults/main.yml
@@ -13,7 +13,7 @@ transmission_group_id: "0"
 
 # network
 transmission_webui_port: "9092"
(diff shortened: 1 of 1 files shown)


davestephens:
The external port on the Ansible-NAS box can be anything. I think you're misunderstanding how port mappings work from your router -> ansible-nas -> transmission docker container...

PR #675, opened 2023-09-20
Update Komga port 25600

jaimesBooth:
Change Komga port to match Komga's new default of 25600: https://komga.org/docs/installation/docker

Change Komga port to match Komga's new default of 25600: https://komga.org/docs/installation/docker
Otherwise unable to access Komga's Tomcat web interface

Fixes #

**Any other useful info**:
New default port specified: https://komga.org/docs/installation/docker


davestephens:
Thanks! :+1:

PR #677, opened 2023-09-23
portainer: option to set whielisted ips in traefik, use port from vars

khartahk:
- Adds same ability to whitelist IPs in traefik for portainer as we have it in bitwarden.
- Sets portainer port to use the one defined in the variables.

**Any other useful info**:
Example for bitwarden: https://github.com/davestephens/ansible-nas/blob/62111605753dfa6f140df21a04c905569b404817/roles/bitwarden/tasks/main.yml#L34

### DIFF
--- a/roles/portainer/tasks/main.yml
+++ b/roles/portainer/tasks/main.yml
@@ -28,6 +28,8 @@
           traefik.http.routers.portainer.tls.domains[0].main: "{{ ansible_nas_domain }}"
           traefik.http.routers.portainer.tls.domains[0].sans: "*.{{ ansible_nas_domain }}"
           traefik.http.services.portainer.loadbalancer.server.port: "9443"
+          traefik.http.routers.portainer.middlewares: "portainer-ipwhitelist@docker"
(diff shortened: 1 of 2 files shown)


roles/portainer/tasks/main.yml
```diff
@@ -27,7 +27,9 @@
           traefik.http.routers.portainer.tls.certresolver: "letsencrypt"
           traefik.http.routers.portainer.tls.domains[0].main: "{{ ansible_nas_domain }}"
           traefik.http.routers.portainer.tls.domains[0].sans: "*.{{ ansible_nas_domain }}"
-          traefik.http.services.portainer.loadbalancer.server.port: "9443"
+          traefik.http.services.portainer.loadbalancer.server.port: "{{ portainer_port }}"
```
davestephens, line 30: This is incorrect - the server port should be the port the container listens on, not the port the user chooses to expose it on.

davestephens (changes requested):

roles/portainer/tasks/main.yml
khartahk: Fixed.

davestephens (approved):

PR #705, opened 2024-04-26
Remove deprecated role geerlingguy.docker_arm

gabrielgradinaru:
Removes deprecated role `geerlingguy.docker_arm` (https://github.com/geerlingguy/ansible-role-docker_arm).
Ansible fails on ARM (M1 Apple silicone in my case) when this role is used.

-

**Any other useful info**:
-

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -9,5 +9,4 @@ skip_list:
 exclude_paths:
  - roles/bertvv.samba
  - roles/geerlingguy.docker
- - roles/geerlingguy.docker_arm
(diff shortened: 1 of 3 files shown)


davestephens:
This might fix your specific use case when running from a Mac, but breaks applying to a Raspberry Pi. 

Can you provide more debug information please? What specifically fails? If you follow the guide when raising an issue, that would be a good start.

gabrielgradinaru:
Sorry for the delay, I've added more details to issue #706.

PR #708, opened 2024-05-24
Fix minecraft docker folder creation

gpambrozio:

The minecraft bedrock server folder was being created even when not configured to be installed.

N/A

**Any other useful info**:

Not a big deal but I noticed this weird folder on my server. This solves it.


davestephens (approved):

davestephens:
Oops...thank you! :)

Issue #710, opened 2024-06-07
Wrong Mealie application port: should be 9000


jacobboon:
**Describe the bug**
[The Mealie role exposes the container port 80.](https://github.com/davestephens/ansible-nas/blob/e60c38b16844b28acec40e8ea6985fee1cb79f21/roles/mealie/tasks/main.yml#L19) When the task is run and the container is started, the mealie port does not work (normally: localhost 9925).

According to [the Mealie docs](https://docs.mealie.io/documentation/getting-started/installation/sqlite/) the app actually runs on port 9000.

**Environment**
- Ansible-NAS revision (`git rev-parse --short HEAD`): e60c38b

**Expected behavior**
Enabling Mealie should let the user access the web app.

**Actual behavior**
The web app cannot be reached because the wrong port is exposed ("Safari cannot reach this page" etc.)

**Steps to reproduce**
- Pull the latest image of the mealie docker container.
- Set `mealie_enabled` true in ansible-nas
- Run your playbook (maybe with tag mealie)

**Additional context**
I'll create a PR if that's ok? :)

davestephens:
Fixed by #695

PR #695, opened 2024-01-23
update mealie port for their 1.0.0 release

eniad:
**What this PR does / why we need it**:

Mealie has released their [v1.0.0](https://github.com/mealie-recipes/mealie/releases/tag/v1.0.0). With all the wonderful updates, they also updated the internal port.

**Which issue (if any) this PR fixes**:

I can make an issue if that's valuable, but it didn't seem needed for such a small change.

PR #711, opened 2024-06-07
Set Mealie app port to 9000

jacobboon:
The Mealie app inside the Docker container runs on port 9000, so we should expose that (rather than port 80).

See #710 

And see [the Mealie docs](https://docs.mealie.io/documentation/getting-started/installation/sqlite/) for reference.

### DIFF
--- a/roles/mealie/tasks/main.yml
+++ b/roles/mealie/tasks/main.yml
@@ -16,7 +16,7 @@
         volumes:
           - "{{ mealie_data_directory }}:/app/data:rw"
         ports:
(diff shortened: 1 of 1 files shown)


davestephens (approved):

davestephens:
Looks good - please can you update `traefik.http.services.mealie.loadbalancer.server.port` accordingly, then we're good to merge,

davestephens (changes requested):
Need to update `traefik.http.services.mealie.loadbalancer.server.port` too.

davestephens:
Dupe of #695

Issue #614, opened 2022-11-18
Change piwigo user and group ids


AlanAndre:
**Describe the bug**
Piwigo refuses to start with the root user, which is provided in roles/piwigo/defaults/main.yml

**Environment**
- Ansible-NAS revision (`git rev-parse --short HEAD`): 5886702a
- Ansible version (paste the entire output of `ansible --version` on the machine you run the playbook from): 2.13.4
- Ansible-NAS operating system (`cat /etc/lsb-release` on the Ansible-NAS box): Ubuntu 22.04.1 LTS

**Expected behavior**
Piwigo starts

**Actual behavior**
"please specify user and group other than root" in the logs of the docker container
webpage gives a 502 error

**Steps to reproduce**
run the playbook

amthna:
@AlanAndre thanks so much for documenting a workaround here! Just ran into this bug.

For others having trouble, just changing the two IDs Alan mentions from "0" to "1" did the trick for me to get Piwigo running.

PR #713, opened 2024-06-09
Fix Piwigo startup and docs for #614

darshitpp:
The host name in the docs is incorrect. I managed to make it work as per: https://www.youtube.com/watch?v=hIGibqdSo6c

Also, piwigo refuses to start with the root user. Changed it to `1` for the current user

### DIFF
--- a/roles/piwigo/defaults/main.yml
+++ b/roles/piwigo/defaults/main.yml
@@ -8,8 +8,8 @@ piwigo_data_directory: "{{ docker_home }}/piwigo/data"
 piwigo_photos: "{{ photos_root }}/piwigo"
 
 # uid / gid
-piwigo_user_id: "0"
-piwigo_group_id: "0"
+piwigo_user_id: "1"
+piwigo_group_id: "1"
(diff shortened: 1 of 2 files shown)


roles/piwigo/defaults/main.yml
```diff
@@ -8,8 +8,8 @@ piwigo_data_directory: "{{ docker_home }}/piwigo/data"
 piwigo_photos: "{{ photos_root }}/piwigo"
 
 # uid / gid
-piwigo_user_id: "0"
-piwigo_group_id: "0"
+piwigo_user_id: "1"
```
davestephens, line 11: I don't think uid/gid 1 is the right one for this - we should probably use 1000

davestephens (changes requested):
Needs a minor tweak, nearly good to go.

roles/piwigo/defaults/main.yml
darshitpp: Fixed

davestephens (approved):
LGTM

PR #717, opened 2024-06-15
Add homepage descriptions and entries for Bazarr, Piwigo, PyLoad, and YoutubeDL Material

darshitpp:
I noticed that some of the services that I used, did not list themselves on `Homepage`. Adding the same.

### DIFF
--- a/roles/piwigo/tasks/main.yml
+++ b/roles/piwigo/tasks/main.yml
@@ -62,7 +62,7 @@
           homepage.group: Photos
           homepage.name: Piwigo
           homepage.icon: piwigo.png
-          homepage.href: "http://{{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }}:80"
+          homepage.href: "http://{{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }}:{{ piwigo_port }}"
           homepage.description: Photos client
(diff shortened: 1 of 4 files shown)


roles/piwigo/tasks/main.yml
```diff
@@ -59,6 +59,11 @@
           traefik.http.routers.piwigo.tls.domains[0].main: "{{ ansible_nas_domain }}"
           traefik.http.routers.piwigo.tls.domains[0].sans: "*.{{ ansible_nas_domain }}"
           traefik.http.services.piwigo.loadbalancer.server.port: "80"
+          homepage.group: Photos
+          homepage.name: Piwigo
+          homepage.icon: piwigo.png
+          homepage.href: "http://{{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }}:80"
```
davestephens, line 65: Please replace 80 with `{{ piwigo_port}}`

davestephens (changes requested):
Nearly there, just a minor tweak needed :)

roles/piwigo/tasks/main.yml
darshitpp: Fixed

davestephens:
Awesome, thanks!

PR #719, opened 2024-07-29
Traefik IPWhiteList deprecated - Use IPAllowList Instead

### LINKED ISSUE #703: Traefik IPWhiteList deprecated - IPAllowList
If you're reporting an issue with execution of the Ansible-NAS playbook, please run the playbook with `-vvv`, and ideally isolate the problematic part with `--tags <ansible tag>` where possible.

**Describe the bug**
Traefik has deprecated the IPwhitelist in favour of IPAllowList. I noticed it with Portainer and it also applies to Bitwarden.

**Steps to reproduce**

Use traefik and make portainer or bitwarden available externally.

**Playbook Output**
From traefik logs

```time="2024-03-30T15:03:51Z" level=warning msg="IPWhiteList is deprecated, please use IPAllowList instead." entryPointName=websecure routerName=websecure-portainer@docker```

**Additional context**
I made the following changes, which seemed to resolve the warning I was getting in traefik logs. Not sure if I did it correctly but here it is for what its worth.
For tasks/main.yml - middleware needs to be updated from IPWhiteList to IPAllowList or ip_whitelist to ip_allowlist. Also the label for middleware source range needs to be changed in 2 spots, ipwhitelist to ipallowlist. And lastly the defualts/main.yml, ip_whitelist needs to change to ip_allowlist.

nickjones33:
Updates references from the deprecated IPWhiteList Traefik middleware to use IPAllowlist instead, so that we don't get deprecation warnings from Traefik

Fixes #703

**Any other useful info**:


nickjones33:
I'm not super familiar with the typical github and/or ansible-nas-specific PR process so please let me know if I need to make any changes here, I will monitor this PR and respond to feedback accordingly.

I noticed Issue #703 sitting out there and figured I could make a PR pretty quickly for it

davestephens:
This is a good change, thank you. 

It has been implemented in a way that will break people with no warning though - please can you add a check for the *whitelist variables being set in the relevant roles, and throw an error if so? You'll find equivalent checks for variables in Drone role, if you need something to use as a reference.

davestephens (changes requested):
This is a good change, thank you.

It has been implemented in a way that will break people with no warning though - please can you add a check for the *whitelist variables being set in the relevant roles, and throw an error if so? You'll find equivalent checks for variables in Drone role, if you need something to use as a reference.

nickjones33:
Thanks for the Drone role example, I hope I've followed it as you expected. I set default values in the *whitelist variables of "deprecated" and threw the builtin.fail if the setting had been changed.

nickjones33:
That makes much more sense!

davestephens (approved):

davestephens:
Awesome, thank you! 👍

PR #723, opened 2024-09-10
Fixed wallabag SYMFONY__ENV__DOMAIN_NAME to be the same with wallabag_hostname

Nemoden:

If `wallabag_hostname` configured to anything different than "wallabag", the front-end isn't rendered correctly because the static files are pointing to `wallabag.{{ ansible_nas_domain }}`

n/a

**Any other useful info**:

n/a


davestephens (approved):

davestephens:
Thanks! 🙏🏻

